PR #39, opened 2015-06-19
Added tests for $where

thecalvinchan:
## Compares performance between queries using the query language and the $where operator.

- Comparing a field in a document to a static integer
- $in
- $elemMatch
- Regex
- Querying nested field, match to a static integer

## Checks performance for more complicated queries using $where

- Comparison between two fields in the same document
- Mixed query using query language and $where
- Comparison between two nested fields in the same document

## Document Shape and Size Experimentation

- Query a collection of very large documents with deeply nested fields using both, query language and $where

### DIFF
--- a/testcases/where.js
+++ b/testcases/where.js
@@ -8,13 +8,15 @@ if ( typeof(tests) != "object" ) {
  * Helper function to generate documents in the collection using the 
  * generator function to generate the documents
  */
- function generateDocs(collection, nDocs, generator) {
-    collection.drop();
-    docs = [];
-    for (var i = 0; i < n; i++) {
-      docs.push(generator());
+function generateDocs(nDocs, generator) {
+    return function(collection) {
+        collection.drop();
+        docs = [];
+        for (var i = 0; i < nDocs; i++) {
+          docs.push(generator());
+        }
+        collection.insert(docs, {'ordered': false});
     }
-    collection.insert(docs);
  }
 
 /**
@@ -33,13 +35,13 @@ function increasingXGenerator() {
  * Generates a documents containing an array of 10 
  * random numbers between 0 and 100
  */
- function arrayGenerator() {
+function arrayGenerator() {
     var results = [];
     for (var j = 0; j < 10; j++) {
       results.push(Math.floor(Math.random() * 101));
     }
-    return {"results": numbers};
- }
+    return {"results": results };
+}
 
 /**
  * Generates documents of the form {x: i, y: j}
@@ -51,7 +53,7 @@ function increasingXGenerator() {
     var y = 0;
     return function() {
       var doc = { "x" : x, "y": y};
-      if (y++ > maxYValue) {
+      if (y++ > numY) {
         y = 0;
         x++;
       }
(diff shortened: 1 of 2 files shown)


amidvidy:
can you squash all these commits in to one commit?

testcases/where.js
```diff
@@ -0,0 +1,442 @@
...
+                      collection.insert( { x : i } );
+                  }
+              },
+              ops : [
+                { op : "find", query : { x : 1 } }
+              ] } );
+
+/**
+ * Setup: creates a collection with 4800 documents of the form { x : i }
+ * Test: Finds { x : 1 } using a $where query with ===
```
amidvidy: elaborate that '===' is type-safe equality

testcases/where.js
```diff
@@ -0,0 +1,442 @@
...
+                      collection.insert( { x : i } );
+                  }
+              },
+              ops : [
+                { op : "find", query : { $where : "this.x === 1" } } 
+              ] } );
+
+/**
+ * Setup: creates a collection with 4800 documents of the form { x : i }
+ * Test: Finds { x : 1 } using a $where query with ==
```
amidvidy: elaborate that '==' is weak equality

testcases/where.js
```diff
@@ -0,0 +1,442 @@
...
+
+/**
+ * Setup: creates a collection with 4800 documents of the form { x : i }
+ * Test: Finds all documents with x between 1 and 6 using $in 
+ */
+tests.push( { name : "Where.v1.In.QueryLanguage",
+              tags: ['query','querylanguage'],
+              pre: function( collection ) {
+                  collection.drop();
+                  for ( var i = 0; i < 4800; i++ ) {
```
amidvidy: be consistent about coding style e.g.
` for ( var i = 0; i < 4800; i++ ) {`

vs

`for (var i = 0; i < 4800; i++) {`

I prefer the later.

testcases/where.js
```diff
@@ -0,0 +1,442 @@
...
+              pre: function( collection ) {
+                  collection.drop();
+                  for ( var i = 0; i < 200; i++ ) {
+                    for ( var j = 0; j < 200; j++ ) {
+                      collection.insert( { x : i, y : j } );
+                    }
+                  }
+              },
+              ops : [
+                { op : "find", query : { $where : "this.x == this.y" } }
```
amidvidy: be consistent about $where syntax as well - e.g. passing the function as a string vs as an actual function. I prefer the latter.

testcases/where.js
```diff
@@ -0,0 +1,442 @@
...
+ * random numbers between 0 and 100
+ * Test: Finds all documents with x containing y such that 80 <= y < 85 
+ * using $elemMatch
+ */
+tests.push( { name : "Where.v1.ElemMatch.QueryLanguage",
+              tags: ['query','querylanguage'],
+              pre: function( collection ) {
+                  collection.drop();
+                  for ( var i = 0; i < 4800; i++ ) {
+                    var numbers = [];
```
amidvidy: can you factor out the pre: block from here and ElemMatch.Where into a common utility function?

testcases/where.js
```diff
@@ -0,0 +1,442 @@
...
+ * Setup: Create collection with documents containing 4 character alphabetic permutations
+ * Test: Find document based on Regex Query
+ */
+tests.push( { name : "Where.v1.Regex.QueryLanguage",
+            tags: ['query','querylanguage'],
+            pre: function(collection) {
+                collection.drop();
+                var coll = generatePermutations();
+                collection.insert(coll)
+            },
```
amidvidy: same about refactoring the pre blocks

testcases/where.js
```diff
@@ -0,0 +1,442 @@
...
+
+/*
+ * Setup: Creates a collection of 13 objects, each with 4 nested levels of 13 fields
+ * Test: Find document through match of deeply nested field using $where
+ */
+tests.push( { name : "Where.v1.SimpleNested.Where",
+            tags: ['query', 'where'],
+            pre: function(collection) {
+                collection.drop();
+                var coll = generateNestedCollection();
```
amidvidy: same about refactoring the pre blocks

dalyd:
Could you post using the code review tool? https://wiki.mongodb.com/display/KERNEL/How+We+Use+Code+Review

testcases/where.js
```diff
@@ -0,0 +1,442 @@
...
+    tests = [];
+}
+
+// Queries that can be written in query language and using $where
+
+/**
+ * Setup: creates a collection with 4800 documents of the form { x : i }
+ * Test: Finds { x : 1 } using query language
+ */
+tests.push( { name : "Where.v1.CompareToInt.QueryLanguage",
```
dalyd: Please remove the .v1 in the testnames. We used to have our tests versioned with v1, v2, etc, but we've taken that out of the test files.

PR #41, opened 2015-08-03
CAP-2209 Add geo case with selective filter

bwzhang:


visualzhou:
@rzh , could you please review and merge this patch?

bwzhang:
Here's the new url: https://github.com/mongodb/mongo-perf/pull/42

PR #43, opened 2016-05-11
Use the hostname provided as argument

joeytwiddle:
Although the argument was being parsed before, it was never being used.

This patch uses the argument in the two calls out to the mongo client.

The only problem I can imagine this might cause is if for some reason the default hostname "localhost" does not work on some machines.  (Presumably the user could fix that by passing `--host 127.0.1.1` or whatever.)

### DIFF
--- a/benchrun.py
+++ b/benchrun.py
@@ -150,14 +150,14 @@ def main():
             args.includeFilter = '%'
 
     # Print version info.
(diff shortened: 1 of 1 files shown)


henrikingo:
This fix seems to have been merged some other route. I will close this PR. Thanks.

PR #54, opened 2018-03-23
SERVER-26751 $indexOfArray can optimize for constant array

KevinCybura:


testcases/pipelines.js
```diff
@@ -1063,6 +1063,29 @@ generateTestCase({
...
+      return { _id: i};
+    },
+    pipeline: [
+      {
+        $project: {
+            index: { $indexOfArray: [ largeConstantArrayGenerator(), 700] },
+        }
+      }
+    ],
+    tags: ["indexOfArray"]
```
henrikingo: If you intend for this test to run daily in the waterfall, then you should also add tag "core".

testcases/pipelines.js
```diff
@@ -1063,6 +1063,28 @@ generateTestCase({
...
+generateTestCase({
+    name: "indexOfArray",
+    nDocs: 30000,
+    docGenerator: function emptyDocGenerator(i) {
+      return { _id: i};
+    },
+    pipeline: [
+      {
+        $project: {
+            index: { $indexOfArray: [ largeConstantArrayGenerator(), 700] },
```
cswanson310: It looks like we aren't actually testing what we want to be testing here. Because both arguments to $indexOfArray are constant, the first optimization (in `ExpressionNary::optimize()`) will kick in and the whole $indexOfArray expression will be optimized to a constant value. I think you should fix this by replacing 700 with "$_id".

PR #55, opened 2018-04-26
SERVER-25957 Optimize $filter + $arrayElemAt to avoid scanning entire array

KevinCybura:


testcases/pipelines.js
```diff
@@ -1063,7 +1063,70 @@ generateTestCase({
     ],
 });
 
-
+generateTestCase({
```
henrikingo: Please add comment before each test case to document what it does. (See previous ones for examples.)

It woould be nice to document both the setup and the test.

testcases/pipelines.js
```diff
@@ -1063,7 +1063,70 @@ generateTestCase({
...
+    name: "arrayElemAtOptimized",
+    nDocs: 50,
+    docGenerator: function endFilterDocGenerator(i) {
+      var arr = [];
+      for (var ind = 0; ind < 1024; ind++) {
+        arr.push({ index: ind });
+      }
+      return { _id: i, name: i, idArray: arr };
```
henrikingo: What is the point of the `name` field? Just to have some field that is discarded by the $project? Please explain in the comment I asked for above.

testcases/pipelines.js
```diff
@@ -1063,7 +1063,70 @@ generateTestCase({
...
+          singleElement: {
+            $arrayElemAt: [
+              {
+                $filter: {
+                  input: "$idArray",
+                  as: "id",
+                  cond: { $gte: ["$$id.index", 4] }
+                }
+              },
+              7
```
henrikingo: Purely as discussion... I'm curious to learn more about how this is optimized? If the $filter command gives some benefit above, then why filter for $gte 4 and not $gte 7?

PR #56, opened 2018-05-10
PERF-1454: Move Insert.LargeDocVector to be last in the task - again

dalyd:
Patch build here: https://evergreen.mongodb.com/version/5af213e8e3c33109e3ca2185

### DIFF
--- a/testcases/simple_insert.js
+++ b/testcases/simple_insert.js
@@ -247,20 +247,6 @@ tests.push( { name: "InsertIndexedStringsNonSimpleCollation",
               ] } );
 
 /*
- * Setup:
- * Test: Insert a vector of large documents. Each document contains a long string
- * Notes: Generates the _id field on the client
- *        
- */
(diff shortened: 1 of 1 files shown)


testcases/simple_insert.js
```diff
@@ -333,3 +319,22 @@ tests.push( { name: "Insert.UniqueIndexCompoundReverse",
                         b: { "#SEQ_INT":
                             { seq_id: 0, start: 0, step: 1, unique: true } } } }
               ] } );
+
+/*
+ * Setup:
+ * Test: Insert a vector of large documents. Each document contains a long string
+ * Notes: Generates the _id field on the client. This test should remain the last test in the file.
```
rtimmons: maybe throw a comment in here that the order of this statement matters - so we don't hit the "move it *again*" trap again.

rtimmons: oh I see you already did - in that case I think your'e okay but consider putting a similar comment above the test as well.

PR #65, opened 2019-05-22
PERF-1824 Add microbenchmarks for pipeline-style updates

ksajme:
I'll add the evergreen results as they are printed...

Here's a link to the branch:
https://evergreen.mongodb.com/version/5ce5b30457e85a2ac52c823e

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,170 @@
+Language:        Cpp
+# BasedOnStyle:  Google
+# ---
+AccessModifierOffset: -4
+AlignAfterOpenBracket: Align
+AlignConsecutiveAssignments: false
+AlignConsecutiveDeclarations: false
+AlignEscapedNewlinesLeft: true
+AlignOperands:   false
+AlignTrailingComments: true
+AllowAllParametersOfDeclarationOnNextLine: true
+AllowShortBlocksOnASingleLine: false
+AllowShortCaseLabelsOnASingleLine: false
+AllowShortFunctionsOnASingleLine: Empty
+AllowShortIfStatementsOnASingleLine: false
+AllowShortLoopsOnASingleLine: false
+AlwaysBreakAfterDefinitionReturnType: None
+AlwaysBreakAfterReturnType: None
+AlwaysBreakBeforeMultilineStrings: true
+AlwaysBreakTemplateDeclarations: true
+BinPackArguments: false
+BinPackParameters: false
+BraceWrapping:
+  AfterClass:      false
+  AfterControlStatement: false
+  AfterEnum:       false
+  AfterFunction:   false
+  AfterNamespace:  false
+  AfterObjCDeclaration: false
+  AfterStruct:     false
+  AfterUnion:      false
+  BeforeCatch:     false
+  BeforeElse:      false
+  IndentBraces:    false
+BreakBeforeBinaryOperators: None
+BreakBeforeBraces: Attach
+BreakBeforeTernaryOperators: true
+BreakConstructorInitializersBeforeComma: false
+ColumnLimit:     100
+CommentPragmas:  ''
+ConstructorInitializerAllOnOneLineOrOnePerLine: true
+ConstructorInitializerIndentWidth: 4
+ContinuationIndentWidth: 4
+Cpp11BracedListStyle: true
+DerivePointerAlignment: false
+DisableFormat:   false
+ExperimentalAutoDetectBinPacking: false
+ForEachMacros:   [ foreach, Q_FOREACH, BOOST_FOREACH ]
+IncludeCategories:
+  - Regex:           '.*'
+    Priority:        1
+IndentCaseLabels: true
+IndentWidth:     4
+IndentWrappedFunctionNames: false
+KeepEmptyLinesAtTheStartOfBlocks: true
+MacroBlockBegin: ''
+MacroBlockEnd:   ''
+MaxEmptyLinesToKeep: 2
+NamespaceIndentation: None
+ObjCBlockIndentWidth: 2
+ObjCSpaceAfterProperty: false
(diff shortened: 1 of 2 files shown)


.clang-format
```diff
@@ -0,0 +1,170 @@
+Language:        Cpp
```
cswanson310: I didn't think you would do it! Nice. This is just copied over from mongodb/mongo, correct? Might not be obvious to other reviewers.

testcases/pipeline_update.js
```diff
@@ -0,0 +1,416 @@
+if (typeof(tests) != "object") {
```
cswanson310: Ah you added a new file. That would make sense to match up with the new task we're planning to add in https://jira.mongodb.org/browse/SERVER-40582. I'll rebase on top of yours for the findAndModify ones and add mine here before I push.

testcases/pipeline_update.js
```diff
@@ -0,0 +1,416 @@
...
+        }
+        collection.insert(docs);
+        collection.getDB().getLastError();
+        collection.ensureIndex({x: 1});
+    },
+    ops: [
+        {
+          op: 'update',
+          query: {_id: {'#RAND_INT_PLUS_THREAD': [0, 100]}},
+          update: [{$addFields: {x: {$add: ['$x', 1]}}}],
```
cswanson310: Back when this was drafted we didn't have $set available. Can you update to use $set throughout? I think it's easier to read.

testcases/pipeline_update.js
```diff
@@ -0,0 +1,416 @@
...
+        }
+        collection.insert(docs);
+        collection.getDB().getLastError();
+        collection.ensureIndex({x: 1});
+    },
+    ops: [
+        {
+          op: "update",
+          multi: true,
+          query: {_id: {$gt: 1590, $lt: 1610}},
```
cswanson310: Looks like you borrowed from 'Contended.Medium.Indexed' rather than 'Contended.Hot.Indexed'.

cswanson310: Makes me want to name the last one 'Mild'.

testcases/pipeline_update.js
```diff
@@ -0,0 +1,416 @@
...
+              }
+          }]
+        },
+    ]
+});
+
+/**
+ * Setup: Populate collection with an array of increasing values and size.
+ * Test: Update the array by filtering out even numbers and then adding some back in as one
+ * operation.
```
cswanson310: I like the test from a benchmark perspective, but it is an amusing way to use the database. "Ah yes, gotta put those numbers back in that numbers array now that we've cleaned out the even ones".

testcases/pipeline_update.js
```diff
@@ -0,0 +1,416 @@
...
+          ]
+        },
+    ]
+});
+
+/**
+ * Setup: Populate collection with two types of integers and an array of previous values for one
+ * type of integer with one historical value each.
+ * Test: First move the existing values of each type into the arrays to story history, creating the
+ * array for each type if it doesn't exist already. Then update the integers.
```
cswanson310: I think to make this more realistic we should use objects, and the history should store the date the value was changed. Something more like user address being updated (so it's a sub-object, something a little larger). Then when we push into 'oldAddresses', push something like `{dateChanged: "$$NOW", prevValue: "$address"}`. That won't work just yet so if you end up racing with https://jira.mongodb.org/browse/SERVER-41115 then use 'new Date()' on the client side for testing and replace it with "$$NOW" before pushing.

.clang-format
ksajme: I imagine it wouldn't produce exactly the same output as running eslint. But since there's no automation and linting with either is voluntary, the files can live next to each other for sure.

testcases/pipeline_update.js
ksajme: :+1:

testcases/pipeline_update.js
ksajme: Yes.

testcases/pipeline_update.js
ksajme: As we spoke about in-person, the test is actually copied from the right source.

testcases/pipeline_update.js
ksajme: I was gonna sort the numbers too but we can't unwind.

testcases/pipeline_update.js
cswanson310: Oops!

testcases/pipeline_update.js
cswanson310: If you're interested in a wild ride, check out http://www.kamsky.org/stupid-tricks-with-mongodb/rank-and-dense-rank-in-aggregation for sorting an array using expressions.

testcases/pipeline_update.js
ksajme: Done using $$NOW.

testcases/pipeline_update.js
```diff
@@ -0,0 +1,416 @@
...
+ */
+tests.push({
+    name: "Update.MaintainHistory",
+    tags: ['update', 'pipeline-updates'],
+    pre: function(collection) {
+        collection.drop();
+        var docs = [];
+        for (var i = 0; i < 2400; i++) {
```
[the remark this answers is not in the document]
ksajme: I copied another task that was doing 4800 but it couldn't consistently handle the load so I cut it in half.

ksajme:
Sorry for the forceful pushes. I updated the tags in accordance with what David suggested. I'm not going to bother posting results in here since Charlie has already made pretty graphs.

testcases/pipeline_update.js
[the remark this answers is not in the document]
dalyd: ack

PR #78, opened 2021-07-23
SERVER-58309 Delete logic surrounding the now-deprecated safe mode

jenniferpeshansky:
Due to failures in [this perf patch](https://spruce.mongodb.com/version/60f992f41e2d174cc3663d1d/tasks), I'm deleting all logic relating to safe mode from the mongo-perf repo as well.

### DIFF
--- a/util/utils.js
+++ b/util/utils.js
@@ -306,8 +306,6 @@ function runTest(test, thread, multidb, multicoll, runSeconds, shard, crudOption
 
     // set crud options
     new_ops.forEach(function (z) {
-        //  when true, safe mode calls GLE after every op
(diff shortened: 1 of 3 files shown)


util/utils.js
```diff
@@ -306,8 +306,6 @@ function runTest(test, thread, multidb, multicoll, runSeconds, shard, crudOption
 
     // set crud options
     new_ops.forEach(function (z) {
-        //  when true, safe mode calls GLE after every op
-        z.safe = (crudOptions.safeGLE.toLowerCase() == 'true' ? true : false)
```
dalyd, line 310: Oh, good catch. We don't use it, but we still set it.

dalyd (approved):
LGTM assuming it's happily running now. Thanks for the diligence on this one.

PR #81, opened 2021-10-01
PERF-2565 Added a Micro-benchmark for getMore Commands

mdashti:
This PR adds a micro-benchmark that applies scan and non-selective-filter over a big collection with:
 - different number of docs in the collection
 - different sizes of documents in the collection
 - different batch-sizes

These operations are followed with several `getMore` commands to read the whole result. This benchmark mostly measures the impact of `getMore` commands, as the query processing part is very simple and most of the time is spent on requesting the whole big result-set via several `getMore` commands.

### DIFF
--- a/testcases/simple_bigcollection.js
+++ b/testcases/simple_bigcollection.js
@@ -1,0 +1,196 @@
+if (typeof(tests) !== "object") {
+    tests = [];
+}
+
+(function() {
+    "use strict";
+
+    Random.setRandomSeed(259);
+
+    /**
+     * Sets up a collection and/or a view with the appropriate documents and indexes.
+     *
+     * @param {Number} nDocs - The number of documents to insert into the collection.
+     * @param {function} docGenerator - A function that takes a document number and returns a
+     * document.
+     * @param {Object[]} indexes - A list of index specs to create on the collection.
+     * @param {Object} collectionOptions - Options to use for view/collection creation.
+     */
+    function collectionPopulator(nDocs, indexes, docGenerator, collectionOptions) {
+        return function(collection) {
+            Random.setRandomSeed(259);
+
+            collection.drop();
+
+            var db = collection.getDB();
+
+            var collectionCreationSpec = {create: collection.getName()};
+            assert.commandWorked(
+                db.runCommand(Object.extend(collectionCreationSpec, collectionOptions)));
+            var bulkOp = collection.initializeUnorderedBulkOp();
+            for (var i = 0; i < nDocs; i++) {
+                bulkOp.insert(docGenerator(i));
+            }
+            bulkOp.execute();
+            indexes.forEach(function(indexSpec) {
+                assert.commandWorked(collection.createIndex(indexSpec));
+            });
+        };
+    }
+
+    /**
+     * Rewrites a query op in benchRun format to the equivalent aggregation command op, also in
+     * benchRun format.
+     */
+    function rewriteQueryOpAsAgg(op) {
+        var newOp = {
+            op: "command",
+            ns: "#B_DB",
+            command: {
+                aggregate: "#B_COLL",
+                pipeline: [],
+                cursor: {}
+            }
+        };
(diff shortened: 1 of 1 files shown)


mdashti:
I'd suggest also apply the following diff to [etc/perf.yml](https://github.com/10gen/mongo/blob/master/etc/perf.yml):

```
diff --git a/etc/perf.yml b/etc/perf.yml
index b633fde51ce..6b6323d4da2 100644
--- a/etc/perf.yml
+++ b/etc/perf.yml
@@ -474,6 +474,18 @@ tasks:
              exclude_filter: single_threaded,
              threads: "1 2 4 8",
              read_cmd: 'true'}
+  - name: big_collection
+    depends_on: *_compile
+    commands:
+      - func: f_run_dsi_workload
+        vars:
+          test_control: microbenchmarks
+          test_control_params: |
+            {include_filter_1: query,
+             include_filter_2: getmore,
+             exclude_filter: none,
+             threads: "1 2 4 8",
+             read_cmd: 'true'}
   - name: views-query
     depends_on: *_compile
     commands:
@@ -735,6 +747,7 @@ microbenchmark-buildvariants:
     - name: compile
       distros:
       - rhel62-large
+    - name: big_collection
     - name: genny_scale_InsertRemove
     - name: genny_execution_UserAcquisition
     - name: aggregation_read_commands

```

If you agree, I can create a separate SERVER ticket for that change and create a separate PR for it.

brad-devlugt:
Hey @mdashti thanks for the PR and new workload. I'm substituting in for David as a reviewer from the perf team. Could you please run a few patches with this change + the etc/perf.yml change to see what the overall noise level is like for the new test?  Please reach out if you have any questions!

mdashti:
@brad-devlugt Thanks for accepting to take a look at this PR.

Here are a few evergreen patches:
1. https://spruce.mongodb.com/version/615cb5f22fbabe26cd1ac66a
2. https://spruce.mongodb.com/version/615cb6790305b951efd5a5e6
3. https://spruce.mongodb.com/version/615cb6cb1e2d175df439fe3c
4. https://spruce.mongodb.com/version/615cb6f1e3c3316dc7eaf06f

testcases/simple_bigcollection.js
```diff
@@ -0,0 +1,199 @@
...
+                assert.commandWorked(collection.createIndex(indexSpec));
+            });
+        };
+    }
+
+    /**
+     * Rewrites a query op in benchRun format to the equivalent aggregation command op, also in
+     * benchRun format.
+     */
+    function rewriteQueryOpAsAgg(op) {
```
brad-devlugt, line 45: This looks relatively similar to the function in simple_query.js.  What are your thoughts on trying to re-use that code?

testcases/simple_bigcollection.js
```diff
@@ -0,0 +1,199 @@
...
+                options.nDocs, indexes, options.docs, options.collectionOptions),
+            post: function(collection) {
+                collection.drop();
+            },
+            ops: [rewriteQueryOpAsAgg(options.op)]
+        });
+    }
+
+    function testQuery(testName, nDocs, docSize, op) {
+        var fullName = testName+"_"+nDocs+"_"+docSize+"_"+(op.batchSize ? op.batchSize : "0");
```
brad-devlugt, line 135: For the full name of the test, what are your thoughts on something like:

`BigCollection.Filter (nDocs: 102400, docSize: 4096, batchSize: 256)`

This could make it easier to filter on the trend charts pages, e.g. "docSize: 4096" would show all results for a specific size.

brad-devlugt (approved):
Thanks for the patch runs!  The measurements look like they have a low noise level, so that is great.  I just had a couple thoughts/ideas and overall LGTM.

testcases/simple_bigcollection.js
mdashti: That makes sense. Changed is based on your suggestion in 017abdf

PR #82, opened 2021-10-04
PERF-1106 Fix Queries.IntNonIdFindOne to query by x instead of _id

nzolnierzmdb:


nzolnierzmdb:
I found this in passing and noticed two others who filed tickets for it, so I figured it's a one line change and easy enough to fix. Patch build is here: https://spruce.mongodb.com/version/615b65102fbabe1bb22b35c7/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC, expecting a slight regression since we won't use IDHACK anymore.

testcases/simple_query.js
```diff
@@ -226,7 +226,7 @@ if (typeof(tests) !== "object") {
             return {_id: i};
```
borosaurus, line 226: We should also add a field 'x' here :)

borosaurus (approved):
LGTM with comment

testcases/simple_query.js
nzolnierzmdb: Nice catch, updated the test below as well. Will trigger another patch build and see how much the results differ.

PR #84, opened 2021-10-19
New benchmarks for find and aggregate with group

IrinaYatsenko:
This change adds on the order of 100 new benchmarks many of which synthesize relatively large collections on the fly. I'm not sure whether this is acceptable.

### DIFF
--- a/testcases/pipelines.js
+++ b/testcases/pipelines.js
@@ -273,6 +273,410 @@ generateTestCase({
     pipeline: [{$group: {_id: "$_idMod10", avg: {$avg: "$_id"}}}]
 });
 
+const smallCollectionSize = 100;
+const largeCollectionSize = 100000;
+
+// The intent of testing with small documents is to have small overhead associated with parsing
+// and copying them while having enough fields to run queries with different characteristics
+// such as selectivity, complex expressions, sub-fields and arrays access, etc.
+const smallDoc = function (i) {
+    return {
+        _id: i,
+        a: Random.randInt(10),
+        b: Random.randInt(1000),
+        c: Random.rand() * 100,
+        d: i % 10000,
+        e: {
+            x: Random.randInt(10),
+            y: Random.randInt(1000),
+            z: { u: Random.randInt(100), v: Random.randInt(100) },
+        },
+        f: [Random.randInt(10), Random.randInt(10), Random.randInt(10)],
+    };
+}
+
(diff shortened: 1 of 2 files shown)


testcases/docGenerators.js
```diff
@@ -0,0 +1,91 @@
+const smallCollectionSize = 100;
+const largeCollectionSize = 100000;
+
+// The intent of testing query or aggregation with small documents is to have small overhead 
+// associated with parsing and copying them while having enough fields to run queries with different
+// characteristics such as selectivity, complex expressions, sub-fields and arrays access, etc.
+const smallDoc = function (i) {
```
dalyd, line 7: Please make sure to set the random seed "Random.setRandomSeed(Some Number)" before calling these. Either here or in your test cases. Maybe both just to be careful that nothing is missed.

util/utils.js
```diff
@@ -263,66 +263,21 @@ function CommandTracer(testName, options) {
...
-        //  same as 'writeCmd', but for read commands
-        z.readCmd = (crudOptions.readCmdMode.toLowerCase() == 'true' ? true : false)
-    });
-
-    // setup an environment to pass to the pre and post
-    var env = {
-        threads: thread
-    };
-
     if ("pre" in test) {
```
dalyd, line 281: I feel like pre and post should maybe still happen in the runTest command. I could imagine still wanting to do some setup even if we already have the collection. For example, we might want to build and drop different indexes while staying with the same dataset. I recognize that all the data is being created in 'pre' steps today, so there is no differentiation between data creation and anything else you might do in pre. 

Thoughts?

testcases/docGenerators.js
IrinaYatsenko: `Random.setRandomSeed(258);` is invoked in the methods that take these doc generators as parameters. Is it OK to see multiple times?

util/utils.js
IrinaYatsenko: How would you feel about separating data creating into its own option/step so `pre` can be used for other things (such as indexes)? The current tests can continue using `pre` for both tasks but the new ones will be able to make the distinction...

testcases/docGenerators.js
dalyd: First, thanks for double checking that. 
Second, the last setRandomSeed wins, so it should be okay to set in the file here, where it will be evaluated at load time, and also in the test file or test function itself.

PR #85, opened 2021-10-20
PERF-2484 Add performance tests for $firstN/$lastN accumulators, expr…

mtandrei:
…essions, and window functions

Evergreen: https://evergreen.mongodb.com/version/616f1414d1fe075a84378253

### DIFF
--- a/testcases/pipelines.js
+++ b/testcases/pipelines.js
@@ -301,7 +301,7 @@ function maxNDocGenerator(i) {
  */
 generateTestCase({
     name: "Group.TenGroupsWithMinN",
-    tags: ['>=5.1.0'],
+    tags: ['>=5.2.0'],
     nDocs: 1000,
     docGenerator: minNDocGenerator,
     pipeline: [{$group: {_id: "$_idMod10", minVals: {$minN: {n: 10, output: "$_id"}}}}]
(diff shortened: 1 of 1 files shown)


testcases/pipelines.js
```diff
@@ -328,7 +328,7 @@ generateTestCase({
  *
  * @param: {Boolean} isMin: true if we're generating an array for $minN.
  */
-function generateProjectArray(isMin){
+function generateMinMaxNProjectArray(isMin){
```
mjrb, line 331: [optional] I would prefer if we change the parameter name to `descending` or `isDescending` then keep the original name of this function and not have to make a wrapper function `generateFirstLastNProjectArray()` that appears further down

testcases/pipelines.js
```diff
@@ -401,6 +406,105 @@ generateTestCase({
...
+    tags: ['>=5.1.0'],
+    nDocs: 1000,
+    docGenerator: function basicGroupDocGenerator(i) {
+        return {_id: i, _idMod10: i % 10};
+    },
+    pipeline: [{$group: {_id: "$_idMod10", firstVals: {$firstN: {n: 10, output: "$_id"}}}}]
+});
+
+/**
+ * Test case which splits 1000 documents into 10 groups of 100 and returns the first 10 values in
```
dalyd, line 424: shoudl this be last 10 values?

testcases/pipelines.js
```diff
@@ -401,6 +406,105 @@ generateTestCase({
...
+        return accumulatorNExpressionDocGenerator(i, generateFirstLastNProjectArray());
+    },
+    pipeline: [{$project: {_id: 0, output: {$lastN: {n: 10, output: "$array"}}}}]
+});
+
+/**
+ * Test case which splits 1000 documents into 10 partitions of 100. For each document within the
+ * partition, we take the first 10 documents over a sliding window of at least 10 documents and up
+ * to 21 documents. This window consists of the 10 documents before the current document, the 10
+ * documents after the current document, and the current document itself.
```
dalyd, line 478: This test case makes my head hurt. Am I correct that the first 10 results will be identical, and then it will start shifting? Maybe (optional) add some sample output to help explain.

testcases/pipelines.js
mtandrei: Done.

testcases/pipelines.js
```diff
@@ -301,7 +301,7 @@ function maxNDocGenerator(i) {
  */
 generateTestCase({
     name: "Group.TenGroupsWithMinN",
-    tags: ['>=5.1.0'],
+    tags: ['>=5.2.0'],
```
dalyd, line 304: Haven't these tests already been running? Are you sure you want to bump to 5.2.0 for the pre-existing tests?

testcases/pipelines.js
dalyd: Thanks

dalyd (approved):
LGTM

I had a question about the versions, but otherwise looks good.

testcases/pipelines.js
mtandrei: Yes, the $minN/$maxN tests have been running on 5.1.0. However, when we close the project and enable the feature flag by default, all of these operators will be pinned to FCV 5.2 (assuming the project closes in the 5.2 cycle). So, this is preemptively bumping the versions for $minN/$maxN so that we don't have any issues down the line (if you want, I can revert this change and hold off until https://jira.mongodb.org/browse/SERVER-52247)

PR #90, opened 2021-11-09
PERF-2689 New benchmarks for large datasets with indexes

IrinaYatsenko:
Evergreen runs:
https://spruce.mongodb.com/version/618d50b45623435d06bc1b3e/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC
https://spruce.mongodb.com/version/618d50172fbabe51bb1beeb1/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

### DIFF
--- a/testcases/pipelines.js
+++ b/testcases/pipelines.js
@@ -2343,5 +2343,77 @@ generateTestCaseWithLargeDataset({
         {$group: {_id: "$f", min: {$min: "$av"}, max: {$max: "$av"}}}
     ]
 });
+
+// Indexed plans.
+let dropIndexesAndCaches = function(collection) {
+    collection.dropIndexes();
+    collection.getPlanCache().clear();
+}
+let createIndexes = function(collection, indexes) {
+    indexes.forEach(function(index) {
+        assert.commandWorked(collection.createIndex(index));
+    });
+}
+function generateTestCaseWithLargeDatasetAndIndexes(options) {
+    options.pre = function(collection) {
+        dropIndexesAndCaches(collection);
+        createIndexes(collection, options.indexes);
+    };
+    options.post = dropIndexesAndCaches;
+    generateTestCaseWithLargeDataset(options);
+}
+
+// Grouping on the indexed field with no accumulators creates DISTINCT_SCAN plan with a
+// $groupByDistinctScan stage, that currently isn't being lowered into SBE.
+generateTestCaseWithLargeDatasetAndIndexes({
+    name: "Group.NoAccTopField_DistinctScan_LL10", 
+    docGenerator: largeDoc,
+    indexes: [{"a":1}],
+    pipeline: [{$group: {_id: "$a"}}]
+});
+
+// The $match stage triggers IXSCAN plans, without it, the only considered plan is COLLSCAN.
+generateTestCaseWithLargeDatasetAndIndexes({
+    name: "Group.NoAccTopField_SingleIndex_LL10",
+    docGenerator: largeDoc,
+    indexes: [{"b":1}],
+    pipeline: [{$match: {b: 42}}, {$group: {_id: "$a"}}]
+});
+generateTestCaseWithLargeDatasetAndIndexes({
+    name: "Group.NoAccTopField_MultipleIndexes_LL10",
+    docGenerator: largeDoc,
+    indexes: [{"b":1}, {"b":1, "c":1}],
+    pipeline: [{$match: {b: 42}}, {$group: {_id: "$a"}}]
+});
+generateTestCaseWithLargeDatasetAndIndexes({
+    name: "Group.SumAccTopField_SingleIndex_LL10",
+    docGenerator: largeDoc,
+    indexes: [{"b":1}],
+    pipeline: [{$match: {b: 42}}, {$group: {_id: "$a", res: {$sum: "$c"}}}]
+});
+generateTestCaseWithLargeDatasetAndIndexes({
+    name: "Group.SumAccTopField_MultipleIndexes_LL10",
+    docGenerator: largeDoc,
+    indexes: [{"b":1}, {"b":1, "c":1}],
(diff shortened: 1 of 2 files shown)


testcases/simple_query.js
```diff
@@ -1036,4 +1036,48 @@ if (typeof(tests) !== "object") {
...
+        indexes: [{"a":1}, {"b":1}, {"a":1, "b":1}],
+        query: {"a": 7, "b":742}
+    });
+    addTestCaseWithLargeDatasetAndIndexes({
+        name: "PointQuerySubField_SingleIndex_LL", 
+        docGenerator: largeDoc,
+        indexes: [{"e.a":1}],
+        query: {"e.a": 7}
+    });
+    addTestCaseWithLargeDatasetAndIndexes({
```
jimoleary, line 1077: Very much an edge case but do we also want an embedded document index test?

```
addTestCaseWithLargeDatasetAndIndexes({
        name: "PointQuerySubFields_EmbeddedMultipleIndexes_LL", 
        docGenerator: largeDoc,
        indexes: [{"e.a":1}, {"e.b":1}, {"e":1}],
        query: {e:{a: 7, b:742}}
    });
```

testcases/pipelines.js
```diff
@@ -2343,5 +2343,77 @@ generateTestCaseWithLargeDataset({
...
+    options.pre = function(collection) {
+        dropIndexesAndCaches(collection);
+        createIndexes(collection, options.indexes);
+    };
+    options.post = dropIndexesAndCaches;
+    generateTestCaseWithLargeDataset(options);
+}
+
+// Grouping on the indexed field with no accumulators creates DISTINCT_SCAN plan with a
+// $groupByDistinctScan stage, that currently isn't being lowered into SBE.
```
jimoleary, line 2367: Should ```lowered``` be ```pushed down```? The scope, design and SERVER-58429 use the term pushdown.

jimoleary (commented):
I have some minor questions but the works looks pretty good.

testcases/simple_query.js
IrinaYatsenko: With these benchmarks I'm trying to pinpoint specific components of the execution engine performance, especially in the context of comparing the old engine to the new one. The "PointQuery_SingleIndex_LL" case covers having IXSCAN & FETCH stages, "PointQuery_MultipleIndexes_LL" triggers multi-planning with the associated overhead and we are curious to know how big that overhead is. And the "SubFields" cases are to provide comparison with the corresponding non-indexed ones. I'm actually considering removing "PointQuerySubFields_MultipleIndexes_LL" because it doesn't provide any additional signal. I believe, the test case with embedded document index falls into the same bucket.

testcases/pipelines.js
```diff
@@ -2343,5 +2343,77 @@ generateTestCaseWithLargeDataset({
...
+generateTestCaseWithLargeDatasetAndIndexes({
```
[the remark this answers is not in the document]
IrinaYatsenko: Hm. Even when there is an index that covers both group_by field and the accumulator field, we generate a COLLSCAN (no indexes are considered). While this is a valid scenario and potentially something to optimize, currently this case would be the same as Group.SumAccTopField_LL10. I'm worried to grow this test suite too much so I'll leave this case out until we specifically target it (the benchmarks being added here are to support the work in SERVER-58429).

IrinaYatsenko (commented):
Replied to @jimoleary comments. Jim, please let me know if you feel strongly about any of them.

testcases/pipelines.js
jimoleary: kk

testcases/pipelines.js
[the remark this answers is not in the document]
jimoleary: makes sesnse, I was just asking for the sake of completeness.

testcases/simple_query.js
jimoleary: understood, it's very much an edge case anyway.

jimoleary (approved):

IrinaYatsenko:
@jimoleary Jim, could you please take another look? I've ran the benchmarks in Evergreen and realized that we are missing signal from low selectivity queries, so I've added them (and will kick off another set of Evergreen runs). Thanks!

testcases/simple_query.js
```diff
@@ -1036,4 +1036,60 @@ if (typeof(tests) !== "object") {
...
+    }
+
+    addTestCaseWithLargeDatasetAndIndexes({
+        name: "PointQuery_SingleIndex_LL", 
+        docGenerator: largeDoc,
+        indexes: [{"a":1}],
+        query: {"a": 7}
+    });
+    addTestCaseWithLargeDatasetAndIndexes({
+        name: "PointQuery_SingleIndex_LowSelectivityMatch_LL",
```
jimoleary, line 1066: nit: shouldn't the name be ```RangeQuery_SingleIndex_LowSelectivityMatch_LL``` (same for all the $gt queries).

jimoleary (approved):
lgtm but left a question about test names.

testcases/pipelines.js
```diff
@@ -2343,5 +2343,73 @@ generateTestCaseWithLargeDataset({
...
+    name: "Group.NoAccTopField_DistinctScan_LL10", 
+    docGenerator: largeDoc,
+    indexes: [{"a":1}],
+    pipeline: [{$group: {_id: "$a"}}]
+});
+
+// The $match stage triggers IXSCAN plans, without it, the only considered plan is COLLSCAN.
+// The filter might considerably reduce the size of the data the aggregation runs on, so we try with
+// filters that have artifitially low selectivity as well.
+generateTestCaseWithLargeDatasetAndIndexes({
```
ericox, line 2378: These look good to me.

ericox (approved):
LGTM from the query side.

PR #92, opened 2021-12-02
PERF-2588 Add performance tests for $sortArray

dobroshynski:
Performance microbenchmarks for new $sortArray expression being added in PM-2552

### DIFF
--- a/testcases/pipelines.js
+++ b/testcases/pipelines.js
@@ -2442,5 +2442,90 @@ generateTestCaseWithLargeDatasetAndIndexes({
     indexes: [{"a":1}],
     pipeline: [{$match: {a: {$gt: 1}}}, {$group: {_id: "$b", res: {$sum: "$c"}}}]
 });
+
+/**
+ * Generates an array of 'numElements' elements to be used for testing the performance of the
+ * $sortArray aggregation expression. Depending on the sort direction, we specifically generate
+ * the array with elements in descending order or ascending order to maximize the number of
+ * comparisons made during expression evaluation to sort the underlying array.
+ *
+ * @param {Number} numElements: size of the array to generate and later sort.
+ * @param {Boolean} isDescending: true if we're generating an array in descending order.
+ * @param {String} variant: controls what type of data the array holds - either 'numbers', 
+ * 'strings', or 'objects'.
+ */
+ function generateArrayForSortArray(numElements, isDescending, variant){
+    let arr = [];
+    for(let idx = 0; idx < numElements; ++idx){
+        let arrVal = isDescending ? -idx : idx;
+        if (variant === "numbers") {
+            arr.push(arrVal);
+        } else if (variant === "strings") {
+            arr.push(arrVal.toString());
+        } else if (variant === "objects") {
+            arr.push({
+                a: { b: { c: arrVal } }
+            });
+        }
+    }
+    return arr;
+}
+
+/**
+ * Function which generates a document containing an array to be used when evaluating $sortArray.
+ */
+function sortArrayDocGenerator(i, arr){
+    return {_id: i, array: arr};
+}
+
+/**
+ * Function to generate a unique test case name when evaluating $sortArray with different number of
+ * elements and sort direction.
+ */
+function buildSortArrayTestCaseName(base, numElements, direction) {
+    const dirString = direction === 1 ? "Asc" : "Desc";
+    return base + "_" + numElements.toString() + "_" + dirString;
+}
+
+/**
+ * A series of generated test cases for evaluating $sortArray using ascending/descending sort
+ * directions and by sorting a varying number of elements using the expression in the pipeline.
+ */
+[-1, 1].forEach(direction => {
+    // In order to maximize the number of comparison operations, we generate the raw array
+    // in ascending order when we are sorting descending (-1) and in descending order if we
+    // are looking to sort in ascending order (1).
+    const shouldGenerateArrayInDescOrder = direction === 1;
+
+    [10, 100].forEach(numElements => {
+        generateTestCase({
+            name: buildSortArrayTestCaseName("Project.SortArray", numElements, direction),
+            tags: ['>=5.2.0'],
+            docGenerator: function generator(i) {
+                return sortArrayDocGenerator(i, generateArrayForSortArray(numElements, shouldGenerateArrayInDescOrder, "numbers"));
+            },
+            pipeline: [{$project: {_id: 0, output: {$sortArray: {input: "$array", sortBy: direction}}}}]
+        });
+
+        generateTestCase({
+            name: buildSortArrayTestCaseName("Project.SortArray_StringArray", numElements, direction),
+            tags: ['>=5.2.0'],
+            docGenerator: function generator(i) {
+                return sortArrayDocGenerator(i, generateArrayForSortArray(numElements, shouldGenerateArrayInDescOrder, "strings"));
+            },
+            pipeline: [{$project: {_id: 0, output: {$sortArray: {input: "$array", sortBy: direction}}}}]
+        });
+
+        // Test cases with arrays of objects.
+        generateTestCase({
+            name: buildSortArrayTestCaseName("Project.SortArray_ObjectArray", numElements, direction),
+            tags: ['>=5.2.0'],
+            docGenerator: function generator(i) {
+                return sortArrayDocGenerator(i, generateArrayForSortArray(numElements, shouldGenerateArrayInDescOrder, "objects"));
+            },
+            pipeline: [{$project: {_id: 0, output: {$sortArray: {input: "$array", sortBy: {"a.b.c": direction}}}}}]
+        });
+    });
+});
 })();
 


testcases/pipelines.js
```diff
@@ -2442,5 +2442,89 @@ generateTestCaseWithLargeDatasetAndIndexes({
     indexes: [{"a":1}],
     pipeline: [{$match: {a: {$gt: 1}}}, {$group: {_id: "$b", res: {$sum: "$c"}}}]
 });
+
+/**
+ * Generates an array of 'numElements' elements to be used for testing the performance of
+ * $sortArray aggregation expression. Depending on the sort direction, we specifically generate
+ * the array with elements in descending order or ascending order to maximizes the number of
```
mtandrei, line 2449: [nit] s/maximizes/maximize

testcases/pipelines.js
```diff
@@ -2442,5 +2442,89 @@ generateTestCaseWithLargeDatasetAndIndexes({
...
+ * comparisons made during expression evaluation to sort the underlying array.
+ *
+ * @param {Number} numElements: size of the array to generate and later sort.
+ * @param {Boolean} isDescending: true if we're generating an array in descending order.
+ * @param {String} variant: controls what data the array holds - either 'numbers', 'strings',
```
mtandrei, line 2454: [nit] instead of 'controls what data', how about 'controls the type of data that the array holds'?

testcases/pipelines.js
```diff
@@ -2442,5 +2442,89 @@ generateTestCaseWithLargeDatasetAndIndexes({
     indexes: [{"a":1}],
     pipeline: [{$match: {a: {$gt: 1}}}, {$group: {_id: "$b", res: {$sum: "$c"}}}]
 });
+
+/**
+ * Generates an array of 'numElements' elements to be used for testing the performance of
+ * $sortArray aggregation expression. Depending on the sort direction, we specifically generate
```
mtandrei, line 2448: [nit] s/the performance of $sortArray/the performance of the $sortArray

testcases/pipelines.js
```diff
@@ -2442,5 +2442,89 @@ generateTestCaseWithLargeDatasetAndIndexes({
...
+ * comparisons made during expression evaluation to sort the underlying array.
+ *
+ * @param {Number} numElements: size of the array to generate and later sort.
+ * @param {Boolean} isDescending: true if we're generating an array in descending order.
+ * @param {String} variant: controls what data the array holds - either 'numbers', 'strings',
+ * or 'objects'.
+ */
+ function generateArrayForSortArray(numElements, isDescending, variant){
+    let arr = [];
+    for(let idx = 0; idx < numElements; ++idx){
```
mtandrei, line 2459: suggest pulling out `isDescending ? -idx : idx` from lines 2461, 2463, 2466 out of the if else block, assigning it to a variable and then using it on those lines. Something like 
`const val = isDescending ? -idx : idx;
...
arr.push(val);
...
arr.push(val.toString);`

and so on

testcases/pipelines.js
```diff
@@ -2442,5 +2442,89 @@ generateTestCaseWithLargeDatasetAndIndexes({
...
+ * directions and by sorting a varying number of elements using the expression in the pipeline.
+ */
+[-1, 1].forEach(direction => {
+    // In order to maximize the number of comparison operations, we generate the raw array
+    // in ascending order when we are sorting descending (-1) and in descending order if we
+    // are looking to sort in ascending order (1).
+    const shouldGenerateArrayInDescOrder = direction === 1;
+
+    [1, 10, 100, 1000].forEach(numElements => {
+        generateTestCase({
```
mtandrei, line 2500: [optional] I guess we're sticking with the default number of documents (I think it's 500?) because the number of documents doesn't matter, rather, the number of elements per array does, right?

mtandrei (changes requested):
This looks good; nice work Andrii! Minor comments only and a couple of questions.

testcases/pipelines.js
```diff
@@ -2442,5 +2442,89 @@ generateTestCaseWithLargeDatasetAndIndexes({
...
+ * A series of generated test cases for evaluating $sortArray using ascending/descending sort
+ * directions and by sorting a varying number of elements using the expression in the pipeline.
+ */
+[-1, 1].forEach(direction => {
+    // In order to maximize the number of comparison operations, we generate the raw array
+    // in ascending order when we are sorting descending (-1) and in descending order if we
+    // are looking to sort in ascending order (1).
+    const shouldGenerateArrayInDescOrder = direction === 1;
+
+    [1, 10, 100, 1000].forEach(numElements => {
```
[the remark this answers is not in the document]
dalyd: Are all 4 levels telling us something interesting? Fine for exploration, but if the answer is no, let's cut it down.

dalyd (commented):
Not much from me. It looks good and it looks like Mihai has a good handle on the details. 

Could you run a patch build and link it to the PR? Also, could you list out the new test names and matching tasks? I'd like to take a look at that once it's in.

testcases/pipelines.js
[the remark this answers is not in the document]
ksuarz: How about sticking to 10, 100 and 1000?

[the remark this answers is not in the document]
dobroshynski: This sounds good, I trimmed it down to just 10, 100, 1000

testcases/pipelines.js
dobroshynski: Done

testcases/pipelines.js
dobroshynski: Done

testcases/pipelines.js
dobroshynski: Done

testcases/pipelines.js
dobroshynski: Done!

testcases/pipelines.js
dobroshynski: Yep I think that's exactly the idea, hence I left it blank to stick with the default 500

dobroshynski:
Thanks a lot for taking a look Mihai & David! Made all the suggested edits and kicked off a patch build [here](https://spruce.mongodb.com/version/61aa62b7562343496401357b/changes). I configured it with a diff to `dsi` to include the feature flag for this expression in that variant, so I hope that should work

dobroshynski:
hey team! now that skunkworks is done, would appreciate a re-review, thanks!

testcases/pipelines.js
mtandrei: nice; is 'arrVal' declared with 'let' because of line 2464? otherwise, recommend making this const.

testcases/pipelines.js
```diff
@@ -2442,5 +2442,90 @@ generateTestCaseWithLargeDatasetAndIndexes({
...
+    return {_id: i, array: arr};
+}
+
+/**
+ * Function to generate a unique test case name when evaluating $sortArray with different number of
+ * elements and sort direction.
+ */
+function buildSortArrayTestCaseName(base, numElements, direction) {
+    const dirString = direction === 1 ? "Asc" : "Desc";
+    return base + "_" + numElements.toString() + "_" + dirString;
```
mtandrei, line 2487: [question] do other workloads in the microbenchmarks use the underscore convention? If no, recommend deleting them.

mtandrei (approved):
LGTM mod a couple last nits; nice work Andrii!

This patch generates 30 test cases, but I think that should be ok given that the run time for a task doesn't go up by that much (I will let Dave determine whether this is acceptable).

testcases/pipelines.js
ksuarz: I think we prefer dots?

testcases/pipelines.js
dobroshynski: Pretty sure I pushed the change just this thing is outdated?

testcases/pipelines.js
dobroshynski: I see `Group.SumAccTopField_SingleIndex_LowSelectivityMatch_LL` and `SumAccTopField_SingleIndex_LL` and more above in the file

dalyd:
> hey david, thanks for the response. I'm not sure I totally follow about the `aggregation_read_commands` taking longer in waterfall than here, it looked to me like everything ran as expected, though since this is my first performance related task, perhaps I am forgetting something. Regarding tests and the number, I did not see < 1 ops/s but perhaps then we can get rid of the `1000` "level" since as you pointed out the levels don't really tell anything particularly interesting. and even with just 10, 100 there should be enough tests given that the code generates sorts for strings, numbers, and objects

I may be looking at the wrong things :) I was having a little bit of trouble figuring out which tasks to look at in your patch build. It looked like some of the new tests were in the aggregation_read_commands task. I'm in the office tomorrow, so the shortest path may be to look over someone's shoulder tomorrow (if someone else is in). 

In the meantime, here's the aggregation_read_commands task in your patch on all feature flags: https://spruce.mongodb.com/task/performance_linux_wt_standalone_all_feature_flags_aggregation_read_commands_patch_c379649913d701e406bec73b51a85ea6b618f677_61aa62b7562343496401357b_21_12_03_18_34_18/tests?execution=0&sortBy=STATUS&sortDir=ASC

Ah phooey -- human error on my part on the times. I see now I must have looked at the wrong task for comparison (agg-query-comparison-read-commands, rather than aggregation-read-commands). Sorry about that bit. 

I do still see Aggregation.Project.SortArray_ObjectArray_1000_Asc reporting 0.4636431077758186 as a result. So, that's one case less than 1. It is the 1000 level. What do Kyle and Mihai think about that one.

testcases/pipelines.js
ksuarz: 🤷

ksuarz:
@dalyd, @dobroshynski -- it is a little sad that the perf is kind of slow that a 1k array can't be sorted... but I also think it's a bit of an extreme case. I'd be okay with dropping it?

The next slowest after the 1k array is `Aggregation.Project.SortArray_ObjectArray_100_Desc`, which is at 6.914469750161532. I'm assuming that's in the stable zone?

dalyd:
@ksuarz that sounds fine to me.

ksuarz (approved):
Based on the latest exchange, this LGTM mod getting rid of the 1k sort case.

dobroshynski:
thanks for the review & comments all. @dalyd  David if you do not have any further objections / requests I will merge this in

dalyd (approved):

PR #94, opened 2021-12-27
PERF-2741 make syntax in testcases/pipelines.js compatible with older shell versions

mjrb:
I manually checked the syntax using this command `echo -e 'load("util/utils.js"); load("util/docGenerators.js");\n' $(ls testcases/ | awk '{print "load(\"testcases/" $1 "\");"}') |  /path/to/mongo/for/version` with 4.0.27 and 3.6.23 shells downloaded from https://www.mongodb.com/try/download/community. The command just loads dependencies and then loads all of the files under testcases to make sure there is no syntax errors on loading.

[evergreen run](https://spruce.mongodb.com/version/61ca129d0305b92df5e4751e/tasks)

bizarrely Array's forEach method works fine in the 4.0 shell but `for (const var of [])` doesn't?


apodelko (approved):
LGTM

PR #101, opened 2022-09-14
PERF-3282 remove unsupported use of  in wildcard benchmarks

mjrb:
I tested this locally and the error went away and it started reporting data. [here](https://spruce.mongodb.com/version/632206249ccd4e77c0d24dc7/tasks) is a patch build that will hopefully eventually finish. When I looked at the benchRun parser it had the option for specifying sort at the top level so I just used that instead.

### DIFF
--- a/testcases/wildcard_index_query.js
+++ b/testcases/wildcard_index_query.js
@@ -240,7 +240,7 @@ if ((typeof tests === "undefined" ? "undefined" : typeof(tests)) != "object") {
             var rangeStart = Random.randInt(maxValue - 10);
 
             query[fieldList[i]] = {$gte: rangeStart, $lte: (rangeStart + 10)};
-            list.push({op: "find", query: {$query: query, $orderby: sort}});
+            list.push({op: "find", query, sort});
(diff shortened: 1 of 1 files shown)


dstorch (approved):
I'm glad it was this simple. LGTM!

I had entirely forgotten, but it appears I was somewhat involved in the discussion on the ticket that Ian implemented to add the `sort` option as a first class citizen in `benchRun()`: [SERVER-5722](https://jira.mongodb.org/browse/SERVER-5722). The mongo-perf code has to work against older versions of the shell that still support OP_QUERY. Luckily, it looks like Ian backported the necessary shell changes all the way back to 3.6. In older versions of the shell, we handle the OP_QUERY legacy reads case by converting the `sort` option to `$orderby` [here](https://github.com/mongodb/mongo/blob/97ba88b52784d3c81a23a2994f50d16f3bf2dab0/src/mongo/shell/bench.cpp#L136-L144).

Anyway, enough of my ramblings. Thanks for noticing this issue and doing the fix!

PR #102, opened 2022-09-23
SERVER-66511: Add benchmark for sort+limit operations

albymassari:


testcases/simple_query.js
```diff
@@ -846,6 +846,299 @@ if (typeof(tests) !== "object") {
         }
     });
 
+    /**
+     * Setup: Create a collection with scalar fields x, y with no index.
+     *
+     * Test: Sort the collection by one or two fields, with and without a limit.
+     */
+    addTestCase({
```
mtandrei, line 854: The 9 tests below (from 849 to 997) are identical save for the number of documents and the limit. It might be more readable if you generate [these using a nested loop ](https://github.com/mongodb/mongo-perf/blob/bfb260f9680b7e61ab21aecdfc5152220c20f8ea/testcases/pipelines.js#L633-L639) like so:

```
for (limit of [null, 1, 100]){
     const kBaseNDocs = 1000;
     for(multiple of [1, 10, 100]){
            const kNumDocs = multiple * kBaseNDocs;
            addTestCase(...);
     }
}
```

Optionally, you could extend this to have 3 levels of nesting, and add:
```
for (const sortKey of [{y: 1}, {y: 1, x: 1})
```
At the very top

This will require dynamically generating the test names like so: https://github.com/mongodb/mongo-perf/blob/bfb260f9680b7e61ab21aecdfc5152220c20f8ea/testcases/pipelines.js#L380

testcases/simple_query.js
```diff
@@ -846,6 +846,299 @@ if (typeof(tests) !== "object") {
         }
     });
 
+    /**
+     * Setup: Create a collection with scalar fields x, y with no index.
+     *
+     * Test: Sort the collection by one or two fields, with and without a limit.
```
mtandrei, line 852: [nit] I would also mention in this comment that we're scaling with an increasing number of documents.

mtandrei (changes requested):
This looks good Alberto! Two comments of substance and this should be good to go.

testcases/simple_query.js
```diff
@@ -846,6 +846,299 @@ if (typeof(tests) !== "object") {
         }
```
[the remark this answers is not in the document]
albymassari: Sure: it's at https://github.com/10gen/PerformanceReports/pull/5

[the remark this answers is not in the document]
mtandrei: Thanks! I'll take a look at it shortly.

testcases/simple_query.js
```diff
@@ -846,6 +846,40 @@ if (typeof(tests) !== "object") {
...
 
+    /**
+     * Setup: Create a collection with scalar fields x, y with no index.
+     *
+     * Test: Sort the collection by one or two fields, with simple and dotted paths,
+     * with and without a limit, for increasing number of documents.
+     */
+    for (const limit of [[null, 'NoLimit'], [1, 'LimitOne'], [100, 'LimitHundred']]) {
+        for (const numdocs of [[1000, '1K'], [10000, '10K'], [100000, '100K']]) {
+            for (const sortKey of [[{ y: 1 }, '1Key'], [{ y: 1, x: 1 }, '2Key'], [{"z.w.j": 1}, '1PathKey']]) {
```
mtandrei, line 857: [nit] suggest naming 1PathKey to 1DottedPathKey or 1PathKey3Components or something like that (though, I understand if you don't want to update this to avoid churn between this PR and https://github.com/10gen/PerformanceReports/pull/5)

mtandrei (approved):
LGTM; thanks for adding these Alberto! Please hold off on merging until the other PR is approved and merged.

PR #103, opened 2022-09-26
SERVER-51562: Add tests that use $subtract and $multiply

Fefer-Ivan:
Add ProjectWithArithExpressions to pipelines.js
Add ProjectWithArithExpressions2 that has more operators than ProjectWithArithExpressions to simple_query.js

### DIFF
--- a/testcases/simple_query.js
+++ b/testcases/simple_query.js
@@ -1038,7 +1038,10 @@ if (typeof(tests) !== "object") {
     });
 
     let projectWithArithExpressions2 = {
-        ab: {$subtract: ["$b", "$a"]}, cd: {$multiply: ["$d", "$c"]},
+        an: {$abs: "$a"}, bn: {$mod: ["$b", 17]}, cn: {$floor: "$c"},
+        dl: {$ln: {$add: [{$abs: "$d"}, 1]}},
+        ab: {$add: ["$a", "$b"]}, cd: {$multiply: ["$c", "$d"]},
+        ba: {$subtract: ["$b", "$a"]}, dc: {$divide: ["$d", "$c"]},
     };
     addTestCaseWithLargeDataset({
         name: "ProjectWithArithExpressions2_CollScan_LS",
--- a/testcases/simple_query.js
+++ b/testcases/simple_query.js
@@ -1037,6 +1037,15 @@ if (typeof(tests) !== "object") {
         op: {op: "find", query: {}, filter: projectWithArithExpressions}
     });
 
+    let projectWithSubtract = {
+        ba: {$subtract: ["$b", "$a"]}, cb: {$subtract: ["$c", "$b"]}, dc: {$subtract: ["$d", "$c"]},
+    };
+    addTestCaseWithLargeDataset({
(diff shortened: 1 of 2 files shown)


testcases/simple_query.js
```diff
@@ -1037,6 +1037,15 @@ if (typeof(tests) !== "object") {
         op: {op: "find", query: {}, filter: projectWithArithExpressions}
     });
 
+    let projectWithSubtract = {
```
lriuui0x0, line 1040: Can we not merge this to the previous $projWithArithExpressions? If we don't want to disturb existing coverage, maybe have a `projectWithArithExpressions2` that covers other missing arithmetics?

testcases/simple_query.js
```diff
@@ -1037,6 +1037,18 @@ if (typeof(tests) !== "object") {
         op: {op: "find", query: {}, filter: projectWithArithExpressions}
     });
 
+    let projectWithArithExpressions2 = {
+        an: {$abs: "$a"}, bn: {$mod: ["$b", 17]}, cn: {$floor: "$c"},
+        dl: {$ln: {$add: [{$abs: "$d"}, 1]}},
+        ab: {$add: ["$a", "$b"]}, cd: {$multiply: ["$c", "$d"]},
+        ba: {$subtract: ["$b", "$a"]}, dc: {$divide: ["$d", "$c"]},
+    };
```
albymassari, line 1045: Aren't you just making a testcase ProjectWithArithExpressions2_CollScan_LS that is extending ProjectWithArithExpressions_CollScan_LS? Why keeping the latter, if it is just missing the subtract and divide operations?

Fefer-Ivan: Performance analysis tools look at execution history of each test to determine stuff like z-value and have more confidence in results.

We need to keep the test at least for some time, until we have a history for the new test.

dstorch (commented):
I'm a little unsure about what specific purpose the new proposed benchmarks would serve. In order to merge the implementation of $subtract, we want to test the performance of the SBE execution of $subtract in isolation. Wouldn't the C++ google benchmark be sufficient for this purpose?

The mongo-perf benchmark tests $subtract but also various other arithmetic expressions. I think we should only add it if we think that it will be concretely useful in the work that @laplab is doing to try to optimize sbe_stage_builder_expression.cpp. It seems like maybe for this purpose it more or less duplicates the pre-existing `ProjectWithArithExpressions_CollScan_LS` benchmark? On the other hand, I suppose it couldn't hurt to add as long as we can afford the AWS bill. I'm just thinking that there are a bunch of benchmarks we need to add for SBE perf, so I want to be as scientific as we can with regards to filling specific gaps in coverage.

I'd be curious to hear what @mtandrei and @laplab think about this as well!

mtandrei:
I think I'm on the same page as Dave here; there's no problem adding this benchmark as is, however, the best way to test the performance of $subtract is through the google microbenchmarks.

Fefer-Ivan:
I have added micro-benchmarks for $subtract in this PR: https://github.com/10gen/mongo/pull/7705
Looks like tests in mongo-perf are not needed. Closing thing PR.

PR #105, opened 2022-10-04
PERF-3365 Extend index scan microbenchmarks

leozzx:
[evergreen perf patch](https://spruce.mongodb.com/version/633b653aa4cf47053a6d24df/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC) (base)
[evergreen perf patch](https://spruce.mongodb.com/version/633b63d13627e075845e4368/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC) (with https://github.com/10gen/mongo/pull/7252)

variant | test | thread_level | measurement | value | value_base | percent_base_diff
-- | -- | -- | -- | -- | -- | --
SBE | Queries.RangeQuery_MultipleIndexes_LowSelectivityMatch_LL | 4 | ops_per_sec | 14.57 | 13.72 | 6.18%
SBE | Queries.RangeQuery_MultipleIndexes_LowSelectivityMatch_LL | 1 | ops_per_sec | 3.82 | 3.66 | 4.29%
Classic | Queries.RangeQuery_MultipleIndexes_LowSelectivityMatch_LL | 4 | ops_per_sec | 16.84 | 14.97 | 12.47%
Classic | Queries.RangeQuery_MultipleIndexes_LowSelectivityMatch_LL | 1 | ops_per_sec | 4.41 | 3.96 | 11.32%
  |   |   |   |   |   |  
SBE | Queries.RangeQuery_MultipleIndexes_LowSelectivityMatch_LS | 4 | ops_per_sec | 14.58 | 13.92 | 4.71%
SBE | Queries.RangeQuery_MultipleIndexes_LowSelectivityMatch_LS | 1 | ops_per_sec | 3.82 | 3.66 | 4.20%
Classic | Queries.RangeQuery_MultipleIndexes_LowSelectivityMatch_LS | 4 | ops_per_sec | 16.77 | 14.87 | 12.76%
Classic | Queries.RangeQuery_MultipleIndexes_LowSelectivityMatch_LS | 1 | ops_per_sec | 4.44 | 3.95 | 12.27%
  |   |   |   |   |   |  
SBE | Queries.RangeQuery_SingleIndex_GenericPlan_LS | 4 | ops_per_sec | 11.65 | 10.89 | 6.97%
SBE | Queries.RangeQuery_SingleIndex_GenericPlan_LS | 1 | ops_per_sec | 3.14 | 2.97 | 5.41%
Classic | Queries.RangeQuery_SingleIndex_GenericPlan_LS | 4 | ops_per_sec | 13.29 | 11.98 | 10.97%
Classic | Queries.RangeQuery_SingleIndex_GenericPlan_LS | 1 | ops_per_sec | 3.61 | 3.30 | 9.54%
  |   |   |   |   |   |  
SBE | Queries.RangeQuery_SingleIndex_GenericPlan_ThreeFields_LS | 4 | ops_per_sec | 10.41 | 7.70 | 35.25%
SBE | Queries.RangeQuery_SingleIndex_GenericPlan_ThreeFields_LS | 1 | ops_per_sec | 2.74 | 2.08 | 32.02%
Classic | Queries.RangeQuery_SingleIndex_GenericPlan_ThreeFields_LS | 4 | ops_per_sec | 11.26 | 10.73 | 4.86%
Classic | Queries.RangeQuery_SingleIndex_GenericPlan_ThreeFields_LS | 1 | ops_per_sec | 2.93 | 2.71 | 8.32%
  |   |   |   |   |   |  
SBE | Queries.RangeQuery_SingleIndex_GenericPlan_TwoFields_LS | 4 | ops_per_sec | 9.73 | 8.15 | 19.39%
SBE | Queries.RangeQuery_SingleIndex_GenericPlan_TwoFields_LS | 1 | ops_per_sec | 2.59 | 2.17 | 19.40%
Classic | Queries.RangeQuery_SingleIndex_GenericPlan_TwoFields_LS | 4 | ops_per_sec | 10.66 | 9.87 | 7.98%
Classic | Queries.RangeQuery_SingleIndex_GenericPlan_TwoFields_LS | 1 | ops_per_sec | 2.86 | 2.64 | 8.30%
  |   |   |   |   |   |  
SBE | Queries.RangeQuery_SingleIndex_GenericPlan_TwoFields_Range_LS | 4 | ops_per_sec | 16.76 | 9.79 | 71.14%
SBE | Queries.RangeQuery_SingleIndex_GenericPlan_TwoFields_Range_LS | 1 | ops_per_sec | 4.30 | 2.64 | 62.89%
Classic | Queries.RangeQuery_SingleIndex_GenericPlan_TwoFields_Range_LS | 4 | ops_per_sec | 17.23 | 15.76 | 9.34%
Classic | Queries.RangeQuery_SingleIndex_GenericPlan_TwoFields_Range_LS | 1 | ops_per_sec | 4.36 | 4.03 | 8.08%
  |   |   |   |   |   |  
SBE | Queries.RangeQuery_SingleIndex_MultiFields_SingleIntervals_LS | 4 | ops_per_sec | 21.34 | 22.49 | -5.14%
SBE | Queries.RangeQuery_SingleIndex_MultiFields_SingleIntervals_LS | 1 | ops_per_sec | 5.67 | 5.94 | -4.52%
Classic | Queries.RangeQuery_SingleIndex_MultiFields_SingleIntervals_LS | 4 | ops_per_sec | 20.39 | 18.40 | 10.84%
Classic | Queries.RangeQuery_SingleIndex_MultiFields_SingleIntervals_LS | 1 | ops_per_sec | 5.47 | 4.91 | 11.35%
  |   |   |   |   |   |  
SBE | Queries.RangeQuery_SingleIndex_SimpleRange_LS | 4 | ops_per_sec | 13.82 | 14.18 | -2.55%
SBE | Queries.RangeQuery_SingleIndex_SimpleRange_LS | 1 | ops_per_sec | 3.70 | 3.84 | -3.58%
Classic | Queries.RangeQuery_SingleIndex_SimpleRange_LS | 4 | ops_per_sec | 14.06 | 12.84 | 9.54%
Classic | Queries.RangeQuery_SingleIndex_SimpleRange_LS | 1 | ops_per_sec | 3.83 | 3.39 | 12.81%
  |   |   |   |   |   |  
SBE | Queries.RangeQuery_SingleIndex_SimpleRange_LowSelectivityMatch_LS | 4 | ops_per_sec | 1,761.92 | 1,774.67 | -0.72%
SBE | Queries.RangeQuery_SingleIndex_SimpleRange_LowSelectivityMatch_LS | 1 | ops_per_sec | 444.59 | 447.13 | -0.57%
Classic | Queries.RangeQuery_SingleIndex_SimpleRange_LowSelectivityMatch_LS | 4 | ops_per_sec | 1,793.84 | 1,567.02 | 14.47%
Classic | Queries.RangeQuery_SingleIndex_SimpleRange_LowSelectivityMatch_LS | 1 | ops_per_sec | 458.40 | 396.62 | 15.58%

### DIFF
--- a/testcases/simple_query.js
+++ b/testcases/simple_query.js
@@ -1135,8 +1135,7 @@ if (typeof(tests) !== "object") {
         query: {"e.a": {$gt: 1}}
     });
 
-    // Select ~1% from a single indexed field. This tests the simple lowKey and highKey range 
-    // IXSCAN plan.
+    // Select ~1% from a single indexed field.
     addTestCaseWithLargeDatasetAndIndexes({
         name: "RangeQuery_SingleIndex_SimpleRange_LowSelectivityMatch_LS",
         tags: ["indexed"],
@@ -1144,8 +1143,8 @@ if (typeof(tests) !== "object") {
         indexes: [{"b": 1}],
         query: {"b": {$gt: 100, $lt: 109}}
     });
-    // Select ~99% from a single indexed field. This tests the simple lowKey and highKey range 
-    // IXSCAN plan.
+
+    // Select ~99% from a single indexed field.
     addTestCaseWithLargeDatasetAndIndexes({
         name: "RangeQuery_SingleIndex_SimpleRange_LS",
         tags: ["indexed"],
@@ -1153,7 +1152,8 @@ if (typeof(tests) !== "object") {
         indexes: [{"b": 1}],
         query: {"b": {$gt: 1, $lt: 999}}
     });
-    // Select ~90% from two indexed field. This tests the optimized single intervals IXSCAN plan.
+
+    // Select ~90% from two indexed field.
     addTestCaseWithLargeDatasetAndIndexes({
         name: "RangeQuery_SingleIndex_MultiFields_SingleIntervals_LS",
         tags: ["indexed"],
@@ -1161,8 +1161,8 @@ if (typeof(tests) !== "object") {
         indexes: [{"b": 1, "h": 1}],
         query: {"b": {"$in": largeArrayRandom}, "h": {$gt: 100}}
     });
-    // Select ~99% from a single indexed field. This tests the generic IXSCAN plan with index 
-    // bounds.
+
+    // Select ~99% from a single indexed field.
     addTestCaseWithLargeDatasetAndIndexes({
         name: "RangeQuery_SingleIndex_GenericPlan_LS",
         tags: ["indexed"],
@@ -1170,7 +1170,8 @@ if (typeof(tests) !== "object") {
         indexes: [{"b": 1}],
         query: {$or: [{"b": {$gt: 99}}, {"b": {$lt: 9}}, {"b": {"$in": largeArrayRandom}}]}
     });
-    // Select ~99% from two indexed field. This tests the generic IXSCAN plan with index bounds.
+
+    // Select ~99% from two indexed field.
     addTestCaseWithLargeDatasetAndIndexes({
         name: "RangeQuery_SingleIndex_GenericPlan_TwoFields_LS",
         tags: ["indexed"],
@@ -1178,7 +1179,8 @@ if (typeof(tests) !== "object") {
         indexes: [{"h": 1, "b": 1}],
         query: {"h": {$gt: 1}, "b": {"$in": largeArrayRandom}}
     });
-    // Select ~99% from three indexed field. This tests the generic IXSCAN plan with index bounds.
+
+    // Select ~99% from three indexed field.
     addTestCaseWithLargeDatasetAndIndexes({
         name: "RangeQuery_SingleIndex_GenericPlan_ThreeFields_LS",
         tags: ["indexed"],
@@ -1186,7 +1188,8 @@ if (typeof(tests) !== "object") {
         indexes: [{"h": 1, "b": 1, "a": 1}],
         query: {"h": {$gt: 1}, "b": {"$in": largeArrayRandom}, "a": {"$in": largeArrayRandom}}
     });
-    // Select ~99% from two indexed field. This tests the generic IXSCAN plan with index bounds.
+
+    // Select ~99% from two indexed field.
     addTestCaseWithLargeDatasetAndIndexes({
         name: "RangeQuery_SingleIndex_GenericPlan_TwoFields_Range_LS",
         tags: ["indexed"],
@@ -1194,7 +1197,8 @@ if (typeof(tests) !== "object") {
         indexes: [{"h": 1, "b": 1}],
         query: {"h": {$gt: 1}, "b": {$lt: 100}}
     });
-    // Select ~99% from three indexed field. This tests the generic IXSCAN plan with index bounds.
+
+    // Select ~99% from three indexed field.
     addTestCaseWithLargeDatasetAndIndexes({
         name: "RangeQuery_SingleIndex_GenericPlan_ThreeFields_Range_LS",
         tags: ["indexed"],
@@ -1202,12 +1206,13 @@ if (typeof(tests) !== "object") {
         indexes: [{"h": 1, "b": 1, "a": 1}],
         query: {"h": {$gt: 1}, "b": {$lt: 100}, "a": {$gt: 1}}
     });
-    // Select ~99% from five indexed field. This tests the generic IXSCAN plan with index bounds.
+
+    // Select ~99% from five indexed field.
     addTestCaseWithLargeDatasetAndIndexes({
         name: "RangeQuery_SingleIndex_GenericPlan_FiveFields_Range_LS",
         tags: ["indexed"],
         docGenerator: smallDoc,
-        indexes: [{"h": 1, "b": 1, "a": 1, "d": 1, "e.a": 1}],
-        query: {"h": {$gt: 1}, "b": {$lt: 100}, "a": {$gt: 1}, "d": {$gt : 10}, "e.a": {$gt: 1}}
+        indexes: [{"h": 1, "b": 1, "e.b": 1, "d": 1, "e.h": 1}],
+        query: {"h": {$gt: 1}, "b": {$lt: 100}, "e.b": {$gt: 1}, "d": {$gt : 10}, "e.h": {$gt: 1}}
     });
 }());
--- a/testcases/simple_query.js
+++ b/testcases/simple_query.js
@@ -1071,8 +1071,8 @@ if (typeof(tests) !== "object") {
     addTestCaseWithLargeDatasetAndIndexes({
         name: "RangeQuery_SingleIndex_MultiFields_LL",
         docGenerator: largeDoc,
-        indexes: [{"a": 1, "b": 1}],
-        query: {"a": {$gt: 1}, "b": {$lt: 900}}
+        indexes: [{"h": 1, "b": 1}],
+        query: {"h": {$gt: 1}, "b": {$lt: 999}}
     });
     addTestCaseWithLargeDatasetAndIndexes({
         name: "PointQuery_MultipleIndexes_LL", 
--- a/testcases/simple_query.js
+++ b/testcases/simple_query.js
@@ -1106,7 +1106,13 @@ if (typeof(tests) !== "object") {
         name: "RangeQuery_SingleIndex_MultiFields_LL",
         docGenerator: largeDoc,
         indexes: [{"h": 1, "b": 1}],
-        query: {"h": {$gt: 1}, "b": {$lt: 999}}
+        query: {"h": {$gt: 1}, "b": {"$in": largeArrayRandom}}
+    });
+    addTestCaseWithLargeDatasetAndIndexes({
+        name: "RangeQuery_SingleIndex_MultiFields2_LL",
+        docGenerator: largeDoc,
+        indexes: [{"h": 1, "b": 1, "a": 1}],
+        query: {"h": {$gt: 1}, "b": {"$in": largeArrayRandom}, "a": {"$in": largeArrayRandom}}
     });
     addTestCaseWithLargeDatasetAndIndexes({
         name: "PointQuery_MultipleIndexes_LL", 
--- a/testcases/simple_query.js
+++ b/testcases/simple_query.js
@@ -1103,18 +1103,6 @@ if (typeof(tests) !== "object") {
         query: {"a": {$gt: 1}}
     });
     addTestCaseWithLargeDatasetAndIndexes({
-        name: "RangeQuery_SingleIndex_MultiFields_LL",
-        docGenerator: largeDoc,
-        indexes: [{"h": 1, "b": 1}],
-        query: {"h": {$gt: 1}, "b": {"$in": largeArrayRandom}}
-    });
-    addTestCaseWithLargeDatasetAndIndexes({
-        name: "RangeQuery_SingleIndex_MultiFields2_LL",
-        docGenerator: largeDoc,
-        indexes: [{"h": 1, "b": 1, "a": 1}],
-        query: {"h": {$gt: 1}, "b": {"$in": largeArrayRandom}, "a": {"$in": largeArrayRandom}}
-    });
-    addTestCaseWithLargeDatasetAndIndexes({
         name: "PointQuery_MultipleIndexes_LL", 
         docGenerator: largeDoc,
         indexes: [{"a":1}, {"b":1}, {"a":1, "b":1}],
@@ -1122,7 +1110,15 @@ if (typeof(tests) !== "object") {
     });
     addTestCaseWithLargeDatasetAndIndexes({
         name: "RangeQuery_MultipleIndexes_LowSelectivityMatch_LL", 
+        tags: ["indexed"],
         docGenerator: largeDoc,
+        indexes: [{"a":1}, {"b":1}, {"a":1, "b":1}],
+        query: {"a": {$gt: 1}, "b": {$lt: 900}}
+    });
+    addTestCaseWithLargeDatasetAndIndexes({
+        name: "RangeQuery_MultipleIndexes_LowSelectivityMatch_LS", 
+        tags: ["indexed"],
+        docGenerator: smallDoc,
         indexes: [{"a":1}, {"b":1}, {"a":1, "b":1}],
         query: {"a": {$gt: 1}, "b": {$lt: 900}}
     });
@@ -1137,5 +1133,55 @@ if (typeof(tests) !== "object") {
         docGenerator: largeDoc,
         indexes: [{"e.a":1}],
         query: {"e.a": {$gt: 1}}
+    });
+
+    addTestCaseWithLargeDatasetAndIndexes({
+        name: "RangeQuery_SingleIndex_SimpleRange_LowSelectivityMatch_LS",
+        tags: ["indexed"],
+        docGenerator: smallDoc,
+        indexes: [{"b": 1}],
+        query: {"b": {$gt: 100, $lt: 109}}
+    });
+    addTestCaseWithLargeDatasetAndIndexes({
+        name: "RangeQuery_SingleIndex_SimpleRange_LS",
+        tags: ["indexed"],
+        docGenerator: smallDoc,
+        indexes: [{"b": 1}],
+        query: {"b": {$gt: 1, $lt: 999}}
+    });
+    addTestCaseWithLargeDatasetAndIndexes({
+        name: "RangeQuery_SingleIndex_MultiFields_SingleIntervals_LS",
+        tags: ["indexed"],
+        docGenerator: smallDoc,
+        indexes: [{"b": 1, "h": 1}],
+        query: {"b": {"$in": largeArrayRandom}, "h": {$gt: 100}}
+    });
+    addTestCaseWithLargeDatasetAndIndexes({
+        name: "RangeQuery_SingleIndex_GenericPlan_LS",
+        tags: ["indexed"],
+        docGenerator: smallDoc,
+        indexes: [{"b": 1}],
+        query: {$or: [{"b": {$gt: 99}}, {"b": {$lt: 9}}, {"b": {"$in": largeArrayRandom}}]}
+    });
+    addTestCaseWithLargeDatasetAndIndexes({
+        name: "RangeQuery_SingleIndex_GenericPlan_TwoFields_LS",
+        tags: ["indexed"],
+        docGenerator: smallDoc,
+        indexes: [{"h": 1, "b": 1}],
+        query: {"h": {$gt: 1}, "b": {"$in": largeArrayRandom}}
+    });
+    addTestCaseWithLargeDatasetAndIndexes({
+        name: "RangeQuery_SingleIndex_GenericPlan_ThreeFields_LS",
+        tags: ["indexed"],
+        docGenerator: smallDoc,
+        indexes: [{"h": 1, "b": 1, "a": 1}],
+        query: {"h": {$gt: 1}, "b": {"$in": largeArrayRandom}, "a": {"$in": largeArrayRandom}}
+    });
+    addTestCaseWithLargeDatasetAndIndexes({
+        name: "RangeQuery_SingleIndex_GenericPlan_TwoFields_Range_LS",
+        tags: ["indexed"],
+        docGenerator: smallDoc,
+        indexes: [{"h": 1, "b": 1}],
+        query: {"h": {$gt: 1}, "b": {$lt: 100}}
     });
 }());
--- a/testcases/simple_query.js
+++ b/testcases/simple_query.js
@@ -1135,6 +1135,8 @@ if (typeof(tests) !== "object") {
         query: {"e.a": {$gt: 1}}
     });
 
+    // Select ~1% from a single indexed field. This tests the simple lowKey and highKey range 
+    // IXSCAN plan.
     addTestCaseWithLargeDatasetAndIndexes({
         name: "RangeQuery_SingleIndex_SimpleRange_LowSelectivityMatch_LS",
         tags: ["indexed"],
@@ -1142,6 +1144,8 @@ if (typeof(tests) !== "object") {
         indexes: [{"b": 1}],
         query: {"b": {$gt: 100, $lt: 109}}
     });
+    // Select ~99% from a single indexed field. This tests the simple lowKey and highKey range 
+    // IXSCAN plan.
     addTestCaseWithLargeDatasetAndIndexes({
         name: "RangeQuery_SingleIndex_SimpleRange_LS",
         tags: ["indexed"],
@@ -1149,6 +1153,7 @@ if (typeof(tests) !== "object") {
         indexes: [{"b": 1}],
         query: {"b": {$gt: 1, $lt: 999}}
     });
+    // Select ~90% from two indexed field. This tests the optimized single intervals IXSCAN plan.
     addTestCaseWithLargeDatasetAndIndexes({
         name: "RangeQuery_SingleIndex_MultiFields_SingleIntervals_LS",
         tags: ["indexed"],
@@ -1156,6 +1161,8 @@ if (typeof(tests) !== "object") {
         indexes: [{"b": 1, "h": 1}],
         query: {"b": {"$in": largeArrayRandom}, "h": {$gt: 100}}
     });
+    // Select ~99% from a single indexed field. This tests the generic IXSCAN plan with index 
+    // bounds.
     addTestCaseWithLargeDatasetAndIndexes({
         name: "RangeQuery_SingleIndex_GenericPlan_LS",
         tags: ["indexed"],
@@ -1163,6 +1170,7 @@ if (typeof(tests) !== "object") {
         indexes: [{"b": 1}],
         query: {$or: [{"b": {$gt: 99}}, {"b": {$lt: 9}}, {"b": {"$in": largeArrayRandom}}]}
     });
+    // Select ~99% from two indexed field. This tests the generic IXSCAN plan with index bounds.
     addTestCaseWithLargeDatasetAndIndexes({
         name: "RangeQuery_SingleIndex_GenericPlan_TwoFields_LS",
         tags: ["indexed"],
@@ -1170,6 +1178,7 @@ if (typeof(tests) !== "object") {
         indexes: [{"h": 1, "b": 1}],
         query: {"h": {$gt: 1}, "b": {"$in": largeArrayRandom}}
     });
+    // Select ~99% from three indexed field. This tests the generic IXSCAN plan with index bounds.
     addTestCaseWithLargeDatasetAndIndexes({
         name: "RangeQuery_SingleIndex_GenericPlan_ThreeFields_LS",
         tags: ["indexed"],
@@ -1177,11 +1186,28 @@ if (typeof(tests) !== "object") {
         indexes: [{"h": 1, "b": 1, "a": 1}],
         query: {"h": {$gt: 1}, "b": {"$in": largeArrayRandom}, "a": {"$in": largeArrayRandom}}
     });
+    // Select ~99% from two indexed field. This tests the generic IXSCAN plan with index bounds.
     addTestCaseWithLargeDatasetAndIndexes({
         name: "RangeQuery_SingleIndex_GenericPlan_TwoFields_Range_LS",
         tags: ["indexed"],
         docGenerator: smallDoc,
         indexes: [{"h": 1, "b": 1}],
         query: {"h": {$gt: 1}, "b": {$lt: 100}}
+    });
+    // Select ~99% from three indexed field. This tests the generic IXSCAN plan with index bounds.
+    addTestCaseWithLargeDatasetAndIndexes({
+        name: "RangeQuery_SingleIndex_GenericPlan_ThreeFields_Range_LS",
+        tags: ["indexed"],
+        docGenerator: smallDoc,
+        indexes: [{"h": 1, "b": 1, "a": 1}],
+        query: {"h": {$gt: 1}, "b": {$lt: 100}, "a": {$gt: 1}}
+    });
+    // Select ~99% from five indexed field. This tests the generic IXSCAN plan with index bounds.
+    addTestCaseWithLargeDatasetAndIndexes({
+        name: "RangeQuery_SingleIndex_GenericPlan_FiveFields_Range_LS",
+        tags: ["indexed"],
+        docGenerator: smallDoc,
+        indexes: [{"h": 1, "b": 1, "a": 1, "d": 1, "e.a": 1}],
+        query: {"h": {$gt: 1}, "b": {$lt: 100}, "a": {$gt: 1}, "d": {$gt : 10}, "e.a": {$gt: 1}}
     });
 }());
--- a/testcases/simple_query.js
+++ b/testcases/simple_query.js
@@ -1148,7 +1148,7 @@ if (typeof(tests) !== "object") {
 
     // Select ~90% with two range predicates on two indexed fields of a compound index.
     addTestCaseWithLargeDatasetAndIndexes({
-        name: "RangeQuery_SingleIndex_MultiFields_SingleIntervals_LS",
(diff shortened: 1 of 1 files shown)


testcases/simple_query.js
```diff
@@ -1110,10 +1110,18 @@ if (typeof(tests) !== "object") {
     });
     addTestCaseWithLargeDatasetAndIndexes({
```
mtandrei, line 1111: This won't generate our favorite flavors of microbenchmarks (that is, 'Queries.IdentityView' and 'Aggregation') because the [underlying function that adds the test only creates ](https://github.com/mongodb/mongo-perf/blob/88d67c08aa58e0319c5be663ca5d53cc3e021674/testcases/simple_query.js#L142)a 'Queries' version, is that right?

testcases/simple_query.js
```diff
@@ -1126,4 +1134,54 @@ if (typeof(tests) !== "object") {
         indexes: [{"e.a":1}],
         query: {"e.a": {$gt: 1}}
     });
+
+    addTestCaseWithLargeDatasetAndIndexes({
+        name: "RangeQuery_SingleIndex_SimpleRange_LowSelectivityMatch_LS",
+        tags: ["indexed"],
+        docGenerator: smallDoc,
+        indexes: [{"b": 1}],
+        query: {"b": {$gt: 100, $lt: 109}}
```
mtandrei, line 1143: Should we note somewhere that this is indeed to match somewhere around 1% of our data set: https://github.com/mongodb/mongo-perf/blob/88d67c08aa58e0319c5be663ca5d53cc3e021674/util/docGenerators.js#L13 
?

testcases/simple_query.js
```diff
@@ -1126,4 +1134,54 @@ if (typeof(tests) !== "object") {
...
+    });
+    addTestCaseWithLargeDatasetAndIndexes({
+        name: "RangeQuery_SingleIndex_SimpleRange_LS",
+        tags: ["indexed"],
+        docGenerator: smallDoc,
+        indexes: [{"b": 1}],
+        query: {"b": {$gt: 1, $lt: 999}}
```
mtandrei, line 1150: Similar comment here, this is going to match ~99% of all docs?

testcases/simple_query.js
```diff
@@ -1126,4 +1134,54 @@ if (typeof(tests) !== "object") {
...
+    });
+    addTestCaseWithLargeDatasetAndIndexes({
+        name: "RangeQuery_SingleIndex_MultiFields_SingleIntervals_LS",
+        tags: ["indexed"],
+        docGenerator: smallDoc,
+        indexes: [{"b": 1, "h": 1}],
+        query: {"b": {"$in": largeArrayRandom}, "h": {$gt: 100}}
```
mtandrei, line 1157: This query seems pretty selective, given that we have 1000 random elements from 0 to 1000: https://github.com/mongodb/mongo-perf/blob/88d67c08aa58e0319c5be663ca5d53cc3e021674/testcases/simple_query.js#L142

testcases/simple_query.js
```diff
@@ -1126,4 +1134,54 @@ if (typeof(tests) !== "object") {
...
+        docGenerator: smallDoc,
+        indexes: [{"b": 1}],
+        query: {$or: [{"b": {$gt: 99}}, {"b": {$lt: 9}}, {"b": {"$in": largeArrayRandom}}]}
+    });
+    addTestCaseWithLargeDatasetAndIndexes({
+        name: "RangeQuery_SingleIndex_GenericPlan_TwoFields_LS",
+        tags: ["indexed"],
+        docGenerator: smallDoc,
+        indexes: [{"h": 1, "b": 1}],
+        query: {"h": {$gt: 1}, "b": {"$in": largeArrayRandom}}
```
mtandrei, line 1171: Should the h predicate be a bit less selective? Or, are we trying to scan as much of the index as possible here?

testcases/simple_query.js
```diff
@@ -1126,4 +1134,54 @@ if (typeof(tests) !== "object") {
...
+        docGenerator: smallDoc,
+        indexes: [{"h": 1, "b": 1, "a": 1}],
+        query: {"h": {$gt: 1}, "b": {"$in": largeArrayRandom}, "a": {"$in": largeArrayRandom}}
+    });
+    addTestCaseWithLargeDatasetAndIndexes({
+        name: "RangeQuery_SingleIndex_GenericPlan_TwoFields_Range_LS",
+        tags: ["indexed"],
+        docGenerator: smallDoc,
+        indexes: [{"h": 1, "b": 1}],
+        query: {"h": {$gt: 1}, "b": {$lt: 100}}
```
mtandrei, line 1185: [idea] We could scale the number of inequalities, so we can see if the "dimensionality" of the index impacts performance. In this test case we have 2, but maybe we could test 3 and 5 fields as well?

mtandrei (commented):
This looks good Zixuan! Some initial comments, I may have another look later in the evening.

testcases/simple_query.js
leozzx: I think so. I didn't see 'Queries.IdentityView' and 'Aggregation' for these test cases.

testcases/simple_query.js
leozzx: let me add comments to each test case.

testcases/simple_query.js
leozzx: Yes, I intend to execute stages we build for ixscan more times.

testcases/simple_query.js
leozzx: Yeah, I want to execute generic ixscan stages as much as possible.

testcases/simple_query.js
```diff
@@ -1126,4 +1134,54 @@ if (typeof(tests) !== "object") {
...
+        docGenerator: smallDoc,
+        indexes: [{"b": 1, "h": 1}],
+        query: {"b": {"$in": largeArrayRandom}, "h": {$gt: 100}}
+    });
+    addTestCaseWithLargeDatasetAndIndexes({
+        name: "RangeQuery_SingleIndex_GenericPlan_LS",
+        tags: ["indexed"],
+        docGenerator: smallDoc,
+        indexes: [{"b": 1}],
+        query: {$or: [{"b": {$gt: 99}}, {"b": {$lt: 9}}, {"b": {"$in": largeArrayRandom}}]}
```
mtandrei, line 1164: In your investigation, the constructed $in arrays are over 100 elements, but in this case, 'largeArrayRandom' is 1000 elements. Should we generate a different, smaller random array for these test cases?

leozzx: Actually larger arrays are preferred to generate larger index bounds and execute the ixscan stage as much as possible. In my investigation, I have to disable the plan cache due to the way I switch between original sbe and my new implementation, so large $in array generate a lot of overhead from array sorting.

testcases/simple_query.js
leozzx: done

testcases/simple_query.js
mtandrei: Thanks; resolving.

testcases/simple_query.js
```diff
@@ -1126,4 +1134,80 @@ if (typeof(tests) !== "object") {
...
+    // Select ~1% from a single indexed field. This tests the simple lowKey and highKey range 
+    // IXSCAN plan.
+    addTestCaseWithLargeDatasetAndIndexes({
+        name: "RangeQuery_SingleIndex_SimpleRange_LowSelectivityMatch_LS",
+        tags: ["indexed"],
+        docGenerator: smallDoc,
+        indexes: [{"b": 1}],
+        query: {"b": {$gt: 100, $lt: 109}}
+    });
+    // Select ~99% from a single indexed field. This tests the simple lowKey and highKey range
```
mtandrei, line 1147: [nit] blank lines in between the end of each test and the subsequent comment for readability

testcases/simple_query.js
mtandrei: Ack; resolving.

testcases/simple_query.js
```diff
@@ -1126,4 +1134,80 @@ if (typeof(tests) !== "object") {
...
+        indexes: [{"h": 1, "b": 1, "a": 1}],
+        query: {"h": {$gt: 1}, "b": {$lt: 100}, "a": {$gt: 1}}
+    });
+    // Select ~99% from five indexed field. This tests the generic IXSCAN plan with index bounds.
+    addTestCaseWithLargeDatasetAndIndexes({
+        name: "RangeQuery_SingleIndex_GenericPlan_FiveFields_Range_LS",
+        tags: ["indexed"],
+        docGenerator: smallDoc,
+        indexes: [{"h": 1, "b": 1, "a": 1, "d": 1, "e.a": 1}],
+        query: {"h": {$gt: 1}, "b": {$lt: 100}, "a": {$gt: 1}, "d": {$gt : 10}, "e.a": {$gt: 1}}
```
mtandrei, line 1211: if we want this to cover ~99% of documents as the comment suggests, shouldn't this be over "e.b" instead of "e.a"? it looks like "e.a" is between 0 and 10, whereas "e.b" is between 0 and 1000:

https://github.com/mongodb/mongo-perf/blob/541d79551a98db66d4ce5088b429f98d7a166e08/util/docGenerators.js#L18

Similar comment for the value of "a", here and in the test case above.

testcases/simple_query.js
mtandrei: Looks good; thanks!

mtandrei (approved):
This LGTM mod a few remaining comments; thanks Zixuan!

testcases/simple_query.js
```diff
@@ -1110,10 +1110,18 @@ if (typeof(tests) !== "object") {
     });
     addTestCaseWithLargeDatasetAndIndexes({
         name: "RangeQuery_MultipleIndexes_LowSelectivityMatch_LL", 
+        tags: ["indexed"],
         docGenerator: largeDoc,
         indexes: [{"a":1}, {"b":1}, {"a":1, "b":1}],
         query: {"a": {$gt: 1}, "b": {$lt: 900}}
     });
+    addTestCaseWithLargeDatasetAndIndexes({
```
dstorch, line 1118: What's the purpose of adding this benchmark? I guess it _could_ use the recursive index scan plan, but only if it selects index `{a: 1, b: 1}`, which doesn't appear to be guaranteed.

testcases/simple_query.js
```diff
@@ -1126,4 +1134,80 @@ if (typeof(tests) !== "object") {
         indexes: [{"e.a":1}],
         query: {"e.a": {$gt: 1}}
     });
+
+    // Select ~1% from a single indexed field. This tests the simple lowKey and highKey range 
+    // IXSCAN plan.
+    addTestCaseWithLargeDatasetAndIndexes({
```
dstorch, line 1140: Do we really not have coverage for something like this? This seems like a simple indexed range scan -- I'd be surprised if we don't have coverage for this already. Here are the tests that we tagged `range_query` in the tags.yml file:

https://github.com/10gen/PerformanceReports/blob/7bec5aa10b8d60ce1e37795a15a198cf579b11f7/scripts/tags.yml#L1516-L1555

testcases/simple_query.js
```diff
@@ -1126,4 +1134,80 @@ if (typeof(tests) !== "object") {
...
+        name: "RangeQuery_SingleIndex_SimpleRange_LowSelectivityMatch_LS",
+        tags: ["indexed"],
+        docGenerator: smallDoc,
+        indexes: [{"b": 1}],
+        query: {"b": {$gt: 100, $lt: 109}}
+    });
+    // Select ~99% from a single indexed field. This tests the simple lowKey and highKey range 
+    // IXSCAN plan.
+    addTestCaseWithLargeDatasetAndIndexes({
```
dstorch, line 1149: Similar here. Isn't something like this already covered?

testcases/simple_query.js
```diff
@@ -1126,4 +1134,80 @@ if (typeof(tests) !== "object") {
...
+        name: "RangeQuery_SingleIndex_SimpleRange_LS",
+        tags: ["indexed"],
+        docGenerator: smallDoc,
+        indexes: [{"b": 1}],
+        query: {"b": {$gt: 1, $lt: 999}}
+    });
+    // Select ~90% from two indexed field. This tests the optimized single intervals IXSCAN plan.
```
dstorch, line 1156: I don't think we should say "This tests the optimized single intervals IXSCAN plan". This comment is tied to the particular's of today's implementation and could easily become stale or confusing in the future. This benchmark will presumably outlive the SBE perf project and will just become one amongst the large corpus of queries that we perf test. Recommend deleting this part of the comment.

testcases/simple_query.js
```diff
@@ -1126,4 +1134,80 @@ if (typeof(tests) !== "object") {
...
+    addTestCaseWithLargeDatasetAndIndexes({
+        name: "RangeQuery_SingleIndex_MultiFields_SingleIntervals_LS",
+        tags: ["indexed"],
+        docGenerator: smallDoc,
+        indexes: [{"b": 1, "h": 1}],
+        query: {"b": {"$in": largeArrayRandom}, "h": {$gt: 100}}
+    });
+    // Select ~99% from a single indexed field. This tests the generic IXSCAN plan with index
```
dstorch, line 1164: Similar here. Recommend deleting "This tests the generic IXSCAN plan with index bounds."

Or if you don't want to delete it you could change it to describe the property of the query rather than the property of the plan that we expect will implement the query.

testcases/simple_query.js
```diff
@@ -1126,4 +1134,80 @@ if (typeof(tests) !== "object") {
...
+        query: {"b": {"$in": largeArrayRandom}, "h": {$gt: 100}}
+    });
+    // Select ~99% from a single indexed field. This tests the generic IXSCAN plan with index 
+    // bounds.
+    addTestCaseWithLargeDatasetAndIndexes({
+        name: "RangeQuery_SingleIndex_GenericPlan_LS",
+        tags: ["indexed"],
+        docGenerator: smallDoc,
+        indexes: [{"b": 1}],
+        query: {$or: [{"b": {$gt: 99}}, {"b": {$lt: 9}}, {"b": {"$in": largeArrayRandom}}]}
```
dstorch, line 1171: What plan do we actually end up generating for this query? Are we smart enough to collapse the branches of the `$or` into the bounds of the index scan.

dstorch (changes requested):
I gave this a quick skim. A few of the proposed added tests don't make sense to me, but overall it looks good.

Do I understand correctly that this _won't_ generate "Aggregation" or "IndentityView" variations of this benchmark. It didn't look like it would, which I think is a good thing, but I wanted to confirm.

testcases/simple_query.js
leozzx: yes, this is same as the above test case `RangeQuery_MultipleIndexes_LowSelectivityMatch_LL` expect that using a `smallDoc` rather than `largeDoc`. I think `largeDoc` have some overhead of document copy and doesn't reflect real index scan performance.

testcases/simple_query.js
leozzx: I didn't check them all, but from their names, there're only three started with `Queries.RangeQuery`, and they don't cover the case that have both low and high bounds.

testcases/simple_query.js
leozzx: Yes, we are able to generate a indexBounds for `$or` on a single field. If there're more than one field, it's going to have a `union` stage for IXSCAN of each field.

leozzx:
> Do I understand correctly that this won't generate "Aggregation" or "IndentityView" variations of this benchmark. It didn't look like it would, which I think is a good thing, but I wanted to confirm.

@dstorch yes, there isn't "Aggregation" or "IndentityView" generated.

leozzx:
Should I also add these tests into [PerformanceReports/tags.yml](https://github.com/10gen/PerformanceReports/blob/main/scripts/tags.yml)? @dstorch

mtandrei (approved):
LGTM; thanks Zixuan!

testcases/simple_query.js
dstorch: What plan does this benchmark actually end up using? And what plan does the `LL` version use?

I'm not sure that the point of the `MultipleIndexes` aspect is. Maybe it's just doing this to try to get the plan cache to be used. Not sure. In any case, it still doesn't entirely make sense to me why we are adding a new benchmark like this.

testcases/simple_query.js
dstorch: It's kind of remarkable that we don't have coverage for a simple index scan with a range bounded on both sides, but it seems you may be right that this was missing. I think there are _id range tests, but I guess it's good to have coverage on a non-_id field.

@mtandrei I'd be curious for your take here. Do you think this is redundant with existing tests, or worth adding?

testcases/simple_query.js
```diff
@@ -1126,4 +1134,85 @@ if (typeof(tests) !== "object") {
         indexes: [{"e.a":1}],
         query: {"e.a": {$gt: 1}}
     });
+
+    // Select ~1% from a single indexed field.
+    addTestCaseWithLargeDatasetAndIndexes({
+        name: "RangeQuery_SingleIndex_SimpleRange_LowSelectivityMatch_LS",
```
dstorch, line 1140: Regarding the name of the test: It looks like elsewhere in this file we are using "LowSelectivityMatch" to mean "unselective", meaning most of the data matches. Is that right? If so, then the naming here seems inconsistent with the rest of the file.

testcases/simple_query.js
dstorch: I guess whatever the answer is for "RangeQuery_SingleIndex_SimpleRange_LowSelectivityMatch_LS" would apply here as well.

testcases/simple_query.js
```diff
@@ -1135,79 +1135,84 @@ if (typeof(tests) !== "object") {
...
     addTestCaseWithLargeDatasetAndIndexes({
         name: "RangeQuery_SingleIndex_SimpleRange_LS",
         tags: ["indexed"],
         docGenerator: smallDoc,
         indexes: [{"b": 1}],
         query: {"b": {$gt: 1, $lt: 999}}
     });
-    // Select ~90% from two indexed field. This tests the optimized single intervals IXSCAN plan.
+
+    // Select ~90% from two indexed field.
```
dstorch, line 1156:
```suggestion
    // Select ~90% with two range predicates on two indexed fields of a compound index.
```

testcases/simple_query.js
dstorch: I'm not sure I understood the second sentence of your response. But you're saying that the plan for this query will be an ixscan with bounds like `{b: [[-Inf, 9), [99, 99], [inval1, inval1], .... [inval_n, inval_n]]}`. Is that right?

testcases/simple_query.js
```diff
@@ -1126,4 +1134,85 @@ if (typeof(tests) !== "object") {
...
+    // Select ~99% from a single indexed field.
+    addTestCaseWithLargeDatasetAndIndexes({
+        name: "RangeQuery_SingleIndex_GenericPlan_LS",
+        tags: ["indexed"],
+        docGenerator: smallDoc,
+        indexes: [{"b": 1}],
+        query: {$or: [{"b": {$gt: 99}}, {"b": {$lt: 9}}, {"b": {"$in": largeArrayRandom}}]}
+    });
+
+    // Select ~99% from two indexed field.
```
dstorch, line 1174:
```suggestion
    // Select ~99% from two indexed fields of a compound index. There is a range predicate on the leading field and a union of point predicates on the trailing field.
```

testcases/simple_query.js
```diff
@@ -1126,4 +1134,85 @@ if (typeof(tests) !== "object") {
...
+    addTestCaseWithLargeDatasetAndIndexes({
+        name: "RangeQuery_SingleIndex_GenericPlan_TwoFields_LS",
+        tags: ["indexed"],
+        docGenerator: smallDoc,
+        indexes: [{"h": 1, "b": 1}],
+        query: {"h": {$gt: 1}, "b": {"$in": largeArrayRandom}}
+    });
+
+    // Select ~99% from three indexed field.
```
dstorch, line 1183:
```suggestion
    // Select ~99% from three indexed fields of a compound index. There is a range predicate on the leading field and unions of point intervals on the trailing fields.
```

testcases/simple_query.js
```diff
@@ -1126,4 +1134,85 @@ if (typeof(tests) !== "object") {
...
+    // Select ~99% from two indexed field.
+    addTestCaseWithLargeDatasetAndIndexes({
+        name: "RangeQuery_SingleIndex_GenericPlan_TwoFields_Range_LS",
+        tags: ["indexed"],
+        docGenerator: smallDoc,
+        indexes: [{"h": 1, "b": 1}],
+        query: {"h": {$gt: 1}, "b": {$lt: 100}}
+    });
+
+    // Select ~99% from three indexed field.
```
dstorch, line 1201:
```suggestion
    // Select ~99% from three indexed fields of a compound index with range predicates on all three fields.
```

testcases/simple_query.js
```diff
@@ -1126,4 +1134,85 @@ if (typeof(tests) !== "object") {
...
+    addTestCaseWithLargeDatasetAndIndexes({
+        name: "RangeQuery_SingleIndex_GenericPlan_ThreeFields_Range_LS",
+        tags: ["indexed"],
+        docGenerator: smallDoc,
+        indexes: [{"h": 1, "b": 1, "a": 1}],
+        query: {"h": {$gt: 1}, "b": {$lt: 100}, "a": {$gt: 1}}
+    });
+
+    // Select ~99% from five indexed field.
```
dstorch, line 1210:
```suggestion
    // Select ~99% from five indexed fields of a compound index.
```

dstorch (changes requested):
Mostly responding to existing threads. Happy to chat about it over Zoom/Slack with you and Mihai if that would be more efficient.

testcases/simple_query.js
leozzx: I believe it's going to use `a_1_b_1` index plan for classic engine, but `a_1` for SBE (will use `a_1_b_1` https://github.com/10gen/mongo/pull/7252 merged). It is same for `LL` version.
I'm going to remove this since it has same perf number as `LL` one in evergreen patch. (it does have huge difference in my local machine)

testcases/simple_query.js
leozzx: Yes.

testcases/simple_query.js
leozzx: Yes, I also find "LowSelectivityMatch" seems be opposite as I understood, I will rename them.

testcases/simple_query.js
mtandrei: Looking at the range_query tests that we do have:
- BigCollection/BigCollectionAggregation don't apply here because[ these are filters over a collection scan](https://github.com/mongodb/mongo-perf/blob/541d79551a98db66d4ce5088b429f98d7a166e08/testcases/simple_bigcollection.js#L92-L108)
- CountsIdRange is what you described, [an id-range query](https://github.com/mongodb/mongo-perf/blob/541d79551a98db66d4ce5088b429f98d7a166e08/testcases/pipelines.js#L1691-L1700) (same with IntIdRange: https://github.com/mongodb/mongo-perf/blob/541d79551a98db66d4ce5088b429f98d7a166e08/testcases/simple_query.js#L236-L244
- CoveredBlockingSort/CoveredNonBlockingSort aren't really about the performance of IXSCANs (at least not directly), similar to IntNonIdFindProjectionCovered and NonCoveredBlockingSortWithIndexToSupportSort
- The PartialIndex test cases arguably don't apply because they're testing partial filter expressions, but even still, they aren't a simple index scan on a single field: https://github.com/mongodb/mongo-perf/blob/541d79551a98db66d4ce5088b429f98d7a166e08/testcases/partial_index.js#L108-L118
- Indeed, the RangeQuery_SingleIndex tests are close, [but not the same](https://github.com/mongodb/mongo-perf/blob/541d79551a98db66d4ce5088b429f98d7a166e08/testcases/simple_query.js#L1099-L1127) (here, a is valued between 1 and 10: https://github.com/mongodb/mongo-perf/blob/541d79551a98db66d4ce5088b429f98d7a166e08/util/docGenerators.js#L54) 
- Finally, the wildcard index case has an index similar to what we have here (in particular, [RangeQueryOnSingleField](https://github.com/mongodb/mongo-perf/blob/541d79551a98db66d4ce5088b429f98d7a166e08/testcases/wildcard_index_query.js#L312-L314); which generates [the following range](https://github.com/mongodb/mongo-perf/blob/541d79551a98db66d4ce5088b429f98d7a166e08/testcases/wildcard_index_query.js#L214) for our field. Arguably though, like the partial index case, this isn't testing a normal B tree index

Given all of that, I think we should go ahead with these test cases (unless we feel that RangeQuery_SingleIndex or the Wildcard "RangeQueryOnSingleField" covers this case already)

dstorch: Ok, given Mihai's analysis above, let's go ahead and add this new test. Thanks for your input @mtandrei! Resolving the thread.

testcases/simple_query.js
```diff
@@ -1126,4 +1127,88 @@ if (typeof(tests) !== "object") {
...
+        name: "RangeQuery_SingleIndex_SimpleRange_LowSelectivityMatch_LS",
+        tags: ["indexed"],
+        docGenerator: smallDoc,
+        indexes: [{"b": 1}],
+        query: {"b": {$gt: 1, $lt: 999}}
+    });
+
+    // Select ~90% with two range predicates on two indexed fields of a compound index.
+    addTestCaseWithLargeDatasetAndIndexes({
+        name: "RangeQuery_SingleIndex_MultiFields_SingleIntervals_LS",
```
dstorch, line 1151:
```suggestion
        name: "RangeQuery_CompoundIndex_SingleIntervals_LS",
```

testcases/simple_query.js
```diff
@@ -1126,4 +1127,88 @@ if (typeof(tests) !== "object") {
...
+        tags: ["indexed"],
+        docGenerator: smallDoc,
+        indexes: [{"b": 1, "h": 1}],
+        query: {"b": {"$in": largeArrayRandom}, "h": {$gt: 100}}
+    });
+
+    // Select ~99% from a single indexed field.
+    addTestCaseWithLargeDatasetAndIndexes({
+        name: "RangeQuery_SingleIndex_GenericPlan_LS",
```
dstorch, line 1160:
```suggestion
        name: "RangeQuery_SingleFieldIndex_ComplexBounds_LS",
```

testcases/simple_query.js
```diff
@@ -1126,4 +1127,88 @@ if (typeof(tests) !== "object") {
...
+        tags: ["indexed"],
+        docGenerator: smallDoc,
+        indexes: [{"b": 1}],
+        query: {$or: [{"b": {$gt: 99}}, {"b": {$lt: 9}}, {"b": {"$in": largeArrayRandom}}]}
+    });
+
+    // Select ~99% from two indexed fields of a compound index. There is a range predicate on the
+    // leading field and a union of point predicates on the trailing field.
+    addTestCaseWithLargeDatasetAndIndexes({
+        name: "RangeQuery_SingleIndex_GenericPlan_TwoFields_LS",
```
dstorch, line 1170:
```suggestion
        name: "RangeQuery_CompoundIndex_ComplexBounds_TwoFields_LS",
```

testcases/simple_query.js
```diff
@@ -1126,4 +1127,88 @@ if (typeof(tests) !== "object") {
...
+        tags: ["indexed"],
+        docGenerator: smallDoc,
+        indexes: [{"h": 1, "b": 1}],
+        query: {"h": {$gt: 1}, "b": {"$in": largeArrayRandom}}
+    });
+
+    // Select ~99% from three indexed fields of a compound index. There is a range predicate on 
+    // the leading field and unions of point intervals on the trailing fields.
+    addTestCaseWithLargeDatasetAndIndexes({
+        name: "RangeQuery_SingleIndex_GenericPlan_ThreeFields_LS",
```
dstorch, line 1180:
```suggestion
        name: "RangeQuery_CompoundIndex_ComplexBounds_ThreeFields_LS",
```

testcases/simple_query.js
```diff
@@ -1126,4 +1127,88 @@ if (typeof(tests) !== "object") {
...
+        tags: ["indexed"],
+        docGenerator: smallDoc,
+        indexes: [{"h": 1, "b": 1, "a": 1}],
+        query: {"h": {$gt: 1}, "b": {"$in": largeArrayRandom}, "a": {"$in": largeArrayRandom}}
+    });
+
+    // Select ~99% from two indexed fields of a compound index with range predicates on both fields.
+    addTestCaseWithLargeDatasetAndIndexes({
+        name: "RangeQuery_SingleIndex_GenericPlan_TwoFields_Range_LS",
```
dstorch, line 1189:
```suggestion
        name: "RangeQuery_CompoundIndex_ComplexBounds_TwoFields_Range_LS",
```

testcases/simple_query.js
```diff
@@ -1126,4 +1127,88 @@ if (typeof(tests) !== "object") {
...
+        tags: ["indexed"],
+        docGenerator: smallDoc,
+        indexes: [{"h": 1, "b": 1}],
+        query: {"h": {$gt: 1}, "b": {$lt: 100}}
+    });
+
+    // Select ~99% from three indexed fields of a compound index with range predicates on all
+    // three fields.
+    addTestCaseWithLargeDatasetAndIndexes({
+        name: "RangeQuery_SingleIndex_GenericPlan_ThreeFields_Range_LS",
```
dstorch, line 1199:
```suggestion
        name: "RangeQuery_CompoundIndex_ComplexBounds_ThreeFields_Range_LS",
```

testcases/simple_query.js
```diff
@@ -1126,4 +1127,88 @@ if (typeof(tests) !== "object") {
...
+        tags: ["indexed"],
+        docGenerator: smallDoc,
+        indexes: [{"h": 1, "b": 1, "a": 1}],
+        query: {"h": {$gt: 1}, "b": {$lt: 100}, "a": {$gt: 1}}
+    });
+
+    // Select ~99% from five indexed fields of a compound index.
+    addTestCaseWithLargeDatasetAndIndexes({
+        name: "RangeQuery_SingleIndex_GenericPlan_FiveFields_Range_LS",
```
dstorch, line 1208:
```suggestion
        name: "RangeQuery_CompoundIndex_ComplexBounds_FiveFields_Range_LS",
```

testcases/simple_query.js
dstorch: > I believe it's going to use a_1_b_1 index plan for classic engine, but a_1 for SBE (will use a_1_b_1 https://github.com/10gen/mongo/pull/7252 merged)

Oh, huh. Do you know why https://github.com/10gen/mongo/pull/7252 causes the change in plan selection? That patch wasn't designed to affect multi-planning in any way, I didn't think.

In any case, I'm in favor of removing this benchmark so I'm happy with the latest changes here. I doubt that these benchmarks really test multi-planning (because of the plan cache), so I'm unsure what the reasoning was behind having multiple indexes in the first place.

Leaving this thread open just to close the loop on the discussion.

dstorch (approved):
LGTM! (with a couple more minor suggestions on naming)

testcases/simple_query.js
leozzx: > Oh, huh. Do you know why https://github.com/10gen/mongo/pull/7252 causes the change in plan selection? That patch wasn't designed to affect multi-planning in any way, I didn't think.

Is it because the plan runs faster after the change? I remember existing SBE uses `a_1` plan but I could be wrong.

dstorch: I doubt that's the explanation. The multi-planner relies on the `numReads` statistic rather than wall clock time. My understanding of https://github.com/10gen/mongo/pull/7252 is that it improves wall clock time but leaves the `numReads` statistic more or less the same. So I didn't expect it to have any obvious impact on the behavior of multi-planning.

PR #107, opened 2022-10-19
PERF-3400 Extend sorting performance tests

leozzx:


testcases/simple_query.js
```diff
@@ -843,6 +843,24 @@ if (typeof(tests) !== "object") {
         op: {op: "find", query: {}}
     });
 
+    /**
+     * Setup: Create a collection with unindexed scalars field 'x'.
+     *
+     * Test: Sort the collection by the 'x' field.
+     */
+    addTestCase({
```
mtandrei, line 851: I think this test case duplicates some of the ones added by Alberto: https://github.com/mongodb/mongo-perf/blob/276696b9804ebe38837091d2c057bd0177af7407/testcases/simple_query.js#L943-L959

Unless there's a good reason to have this test, perhaps this should be deleted

testcases/simple_query.js
```diff
@@ -933,6 +951,104 @@ if (typeof(tests) !== "object") {
         }
     });
 
+    /**
+     * Setup: Create a collection with field 'x' and indexed field 'y'.
+     *
+     * Test: Query the collection by 'x' field and sort by the 'y' field.
+     */
+    addTestCase({
+        name: "NonBlockingSortWithCollScan",
```
mtandrei, line 960: Two things:
- The name is misleading, this isn't a coll scan right? Isn't this a FETCH + IXSCAN?
- This seems to duplicate https://github.com/mongodb/mongo-perf/blob/276696b9804ebe38837091d2c057bd0177af7407/testcases/simple_query.js#L898 ; should we consider deleting this?

testcases/simple_query.js
```diff
@@ -933,6 +951,104 @@ if (typeof(tests) !== "object") {
...
+    });
+
+    /**
+     * Setup: Create a collection with indexed field 'x', and unindexed field 'y'.
+     *
+     * Test: Query the collection by 'x' and 'y' fields using $or then sort by the 'x' field,
+     * MERGE_SORT will be used to merge two branches from $or.
+     */
+    addTestCase({
+        name: "NonCoveredMergeSort",
```
mtandrei, line 1008: Could we also add 'Covered' merge sort tests cases as well?

testcases/simple_query.js
```diff
@@ -933,6 +951,104 @@ if (typeof(tests) !== "object") {
...
+        },
+        indexes: [{y: 1}],
+        op: {
+            op: "find",
+            query: {x: {$gt: 0}},
+            filter: {x: 1, y:1, _id: 0},
+            sort: {y: 1},
+        }
+    });
+
```
mtandrei, line 976: Recommend adding an explode for [sort test case as outlined in the design](https://jira.mongodb.org/browse/PERF-3400). That is, something like this:

```
Query: {a: {$in: [1, 2, …, 9]}}
Sort: {b: 1}
Index: {a: 1, b: 1}
```

This is meant to stress query planning since we don't cache explode for sort plans in the SBE plan cache: https://jira.mongodb.org/browse/SERVER-67576

mtandrei (changes requested):
Looks good Zixuan! Just a couple of comments about potential duplication and some requests for additional tests and this should be good to go.

testcases/simple_query.js
leozzx: Is this what you need? The difference is that both fields are used in the query and it's a range not $in.
https://github.com/mongodb/mongo-perf/blob/276696b9804ebe38837091d2c057bd0177af7407/testcases/simple_query.js#L977-L993

testcases/simple_query.js
leozzx: This is "NonCoveredNonBlockingSort", it's different than CoveredNonBlockingSort because there's a unindexed field used in the query.

testcases/simple_query.js
leozzx: you're right, will delete this one.

leozzx (commented):
@mtandrei There is a comment in other test cases
```
// TODO (SERVER-5722): We cannot create a views passthrough because benchRun doesn't support 
// sorting when running in read command mode. 
createViewsPassthrough: false
```
The SERVER-5722 is fixed already, should I set `createViewsPassthrough` to true?
And how about `createAggregationTest`?

testcases/simple_query.js
mtandrei: Ok that makes sense; then I would rename the test but ignore the second part of my comment.

testcases/simple_query.js
mtandrei: 🤦  my apologies; resolving.

testcases/simple_query.js
mtandrei: Unless I'm mistaken, I don't think that 'NonCoveredBlockingSortWithIndexToSupportSort' exercises the explode for sort path. IIUC, the trick with explode for sort is that we can use a prefix of the index to answer the filter predicate with a series of fixed bounds, but then use bounds on the suffix to answer the sort portion of the query. This results in a series of IXSCANs joined together with a MERGE_SORT stage. In the case of a range query, I'm not sure that this is possible.

testcases/simple_query.js
```diff
@@ -843,6 +843,24 @@ if (typeof(tests) !== "object") {
...
 
+    /**
+     * Setup: Create a collection with unindexed scalars field 'x'.
+     *
+     * Test: Sort the collection by the 'x' field.
+     */
+    addTestCase({
+        name: "SortByScalars",
+        tags: ["core", "sort"],
+        // TODO: SERVER-5722 is fixed, should we set this (and others) to true?
```
mtandrei, line 854: Sorry, I missed this comment on the first pass. Even though SERVER-5722 is fixed, I don't think that it's worth adding these. What will a view or aggregation version of these tests offer that we don't already have in terms of perf coverage?

mtandrei (commented):
Responding to comments; sorry I missed the comment about the views/agg passthrough (TODO SERVER-5722).

testcases/simple_query.js
leozzx: done

testcases/simple_query.js
leozzx: Chatted with Dave, he thinks the views passthroughs are not interesting in our context, but theoretically they might be useful going forward, I'll keep it same as other sorting test cases, we may want to change it in the future.

testcases/simple_query.js
```diff
@@ -933,6 +933,130 @@ if (typeof(tests) !== "object") {
...
+        createViewsPassthrough: false,
+        createAggregationTest: false,
+        nDocs: 1000,
+        docs: function(i) {
+            return {a: Random.randInt(10), b: Random.randInt(10000)};
+        },
+        indexes: [{a: 1, b: 1}],
+        op: {
+            op: "find",
+            query: {a: {$in: [1, 3, 5, 7, 9]}},
```
mtandrei, line 1054: [question] the most ixscans that we can "explode" [before giving up is 200 by default](https://github.com/mongodb/mongo/blob/3edb2e3a7f7320692899c0a295f7d580e38783a3/src/mongo/db/query/query_knobs.idl#L319-L329). Should we have a higher (10 or 20) values within the $in to really stress this logic?

testcases/simple_query.js
mtandrei: Looks good; thanks!

mtandrei (approved):
LGTM; thanks Zixuan!

testcases/simple_query.js
leozzx: done

PR #108, opened 2022-10-20
PERF-3042 Add $group benchmarks for aggregating into a single group

dstorch:
See the comments on the ticket, especially [here](https://jira.mongodb.org/browse/PERF-3042?focusedCommentId=4915583&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-4915583), for additional context.

I ran the queries locally and confirmed that they return apparently sensible results. I also ran the benchmarks locally through `benchrun.py` and got reasonable ops/sec numbers. Also kicked off a [patch build here](https://spruce.mongodb.com/version/635161180305b9742c217f39/tasks).

### DIFF
--- a/testcases/pipelines.js
+++ b/testcases/pipelines.js
@@ -2596,6 +2596,42 @@ generateTestCaseWithLargeDataset({
     ]
 });
 
+// Aggregating the entire collection into a single group.
+generateTestCaseWithLargeDataset({
+    name: "Group.All_SumAccTopField_LL",
+    docGenerator: largeDoc,
+    pipeline: [{$group: {_id: null, res: {$sum: "$c"}}}]
+});
+generateTestCaseWithLargeDataset({
+    name: "Group.All_MultipleAccDiffTopFields_LL",
+    docGenerator: largeDoc,
+    pipeline: [{$group: {_id: null, min: {$min: "$c"}, max: {$max: "$g"}, avg: {$avg: "$h"}}}]
+});
+generateTestCaseWithLargeDataset({
+    name: "Group.All_MultipleAccSameSubFieldStress_LL",
+    docGenerator: largeDoc,
+    pipeline: [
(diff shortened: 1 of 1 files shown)


mtandrei (approved):
LGTM mod some suggestions; thanks Dave!

dstorch (commented):
Replied to Mihai's comments, but no further code changes. I've kicked off two new patch builds in order to see how stable the numbers are:

- Patch 2: https://spruce.mongodb.com/version/6356ada57742ae0bd6e52b59/tasks
- Patch 3: https://spruce.mongodb.com/version/6356ade33066156bda2e1852/tasks

This is now waiting for feedback from David/Mihai regarding whether these tests might be too noisy, as well as the results from these two new patch builds.

dstorch:
The results from my patch builds are in. I put together [this spreadsheet to try to have a preliminary analysis of the noise](https://docs.google.com/spreadsheets/d/1Qlvc5GVpZk7m2et76gzXsiuxsjYO-TEUa1xcy0MLAdI/edit). The good news is that the noise levels look ok to me, even for the tests that have very low ops/sec throughput numbers. I think this means that these changes are good to merge. @dalyd @mtandrei please confirm that you agree with this assessment.

The spreadsheet also shows the SBE vs. classic comparison for the new benchmarks. It looks like we have substantial improvements for 3 out of the 4 benchmarks, with a ~29% regression in `Aggregation.Group.All_MultipleAccDiffSubFieldStress_LL`. This is not a huge surprise, since we have a regression in a few of the other "Stress" benchmarks that this was based upon.

mtandrei (approved):
still LGTM just responding to comments

mtandrei:
To be clear: yes, I agree that the numbers you've linked are fairly stable, so I'm still ok for this to be merged.

PR #110, opened 2022-10-24
PERF-3400 Add testcase for sorting using keys having a common prefix

albymassari:


mtandrei (approved):
LGTM; thanks Alberto!

PR #111, opened 2022-10-28
PERF-3484 Benchmark update operation not requiring index updates

albymassari:
Used to test improvements of SERVER-65364

### DIFF
--- a/testcases/complex_update.js
+++ b/testcases/complex_update.js
@@ -54,10 +54,11 @@ tests.push( { name: "Update.SetWithMultiIndex.Random",
 
 /*
  * Setup: same setup as Update.SetWithMultiIndex.Random, but this time one
- *        of the indexes should not be updated because unaffected by the change.
+ *        of the indexes should not be updated because it is unaffected by
+ *        the change.
  */
 tests.push( { name: "Update.SetWithIgnoredIndex.Random",
-              tags: ['update', 'indexed'],
+              tags: ['update','indexed'],
               pre: function (collection) {
                   collection.drop();
                   var docs = [];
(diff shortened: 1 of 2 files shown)


testcases/complex_update.js
```diff
@@ -77,3 +99,25 @@ tests.push( { name: "Update.SetWithMultiIndex.String",
                     query: { _id : { "#RAND_INT_PLUS_THREAD" : [ 0, 100 ] } },
                     update: { $set : { x : {"#RAND_INT": [0,1000] }, y : {"#RAND_STRING": [1024] } } } },
               ] } );
+
+/*
+ * Setup: same setup as Update.SetWithMultiIndex.String, but this time the
+ *        string index should not be updated because unaffected by the change.
+ */
+tests.push( { name: "Update.SetWithIgnoredIndex.String",
```
banarun, line 107: Can we also add a case where we do update which does `multi:true` update and updates all the documents?

banarun (approved):
LGTM! Thanks @albymassari.

testcases/complex_update.js
albymassari: Sure. I added it in the long_multi_update.js because there was already a test with `multi: true` that had two indexes defined. I just made the query modify only one of the fields, and the results are

```
MultiUpdate.BigAllDocsMultiChange.NoIndex
1       0.12562784083806836
MultiUpdate.BigAllDocsMultiChange.Indexed
1       0.06040132573661832
MultiUpdate.BigAllDocsMultiChangeIgnoredIndex.Indexed
1       0.08830064178672463
```

The "indexed" is touching two fields and requiring the update of both indexes, the "IgnoredIndex" only sets one field and now modifies just one index

PR #113, opened 2022-11-03
Added a note to explain 6.0 change to mongosh

ghartnett:


README.md
```diff
@@ -6,6 +6,8 @@ Mongo-perf (not to be confused with mongoperf) is a micro benchmarking tool for
 This repo contains scripts to run benchmark tests for MongoDB.
 
 ### DEPENDENCIES:
+Since 6.0, the legacy "mongo" shell has been replaced with "mongosh" which does not support benchRun(). To run this micro benchmark with MongoDB 6.0 you'll also need to download the 5.0 legacy "mongo" shell and ensure it's in your path.
```
dalyd, line 9: Should we include some instructions on how to download the legacy shell?

dalyd (approved):
Probably worth a couple more words on how to download the correct thing, but otherwise LGTM.

README.md
dstorch, line 9: I'm not sure that we should document it this way. My vote would be to take an even stronger stance and say that this tool is not user-facing or supported whatsoever for external users by MongoDB. We should make it clear that it's "use at your own risk".

That said, if we do make it clear that we're not supporting this project for external use, then it seems fine to include the hint about the "choose your own adventure" path that involves downloading the 5.0 shell.

dstorch (changes requested):
Thanks for picking this up @ghartnett! I had it on my list to file a ticket about this, but you beat me to the punch 😄 .

I'm curious what you think about the comment I left.

dstorch:
@ghartnett what's the status of this PR? I think it would be good to do this given the recent evidence we've had of external people actually using this repository. I do care about the comment I left being addressed before this goes in, though.

README.md
ghartnett: I'm concerned that more detailed instructions might be brittle and break in future. If you can suggest specific wording that would not be brittle I'll happily add it.

ghartnett:
@dstorch simple unsupported note added near the top. Please let me know if it works for you.

dstorch (approved):
Thanks @ghartnett! This LGTM now. I left a couple additional suggestions which I think are improvements, but I don't need to re-review the resulting changes.

README.md
```diff
@@ -2,10 +2,14 @@
 
 Mongo-perf (not to be confused with mongoperf) is a micro benchmarking tool for the MongoDB server. It measures throughput of commands with regards to the number of threads.
 
+**NOTE: this tool is not supported for users outside MongoDB. Use at your own risk.**
+
 ### OVERVIEW:
 This repo contains scripts to run benchmark tests for MongoDB.
 
 ### DEPENDENCIES:
+Since 6.0, the legacy "mongo" shell has been replaced with "mongosh" which does not support benchRun(). To run this micro benchmark with MongoDB 6.0 you'll also need to download the 5.0 legacy "mongo" shell and ensure it's in your path.
```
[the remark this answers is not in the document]
ghartnett: Done

PR #117, opened 2022-12-13
PERF-3569 Add more realistic scan test cases

lriuui0x0:


testcases/simple_query.js
```diff
@@ -24,6 +24,51 @@ if (typeof(tests) !== "object") {
...
 
+    /**
+     * Setup: Create a collection of documents containing 5 fields other than _id.
+     *
+     * Test: Empty query that returns all documents.
+     */
+    addQueryTestCase({
+        name: "FiveField",
+        tags: ["regression"],
+        nDocs: 100,
```
mtandrei, line 35: Note that we already have such a test case that has a small number of documents: https://github.com/mongodb/mongo-perf/blob/fbc29ee5b451ca73268a9207d55e47f74af74eb9/testcases/simple_query.js#L168-L184

I would say, we want test cases with 10k and 100k documents and then pick either the smallDoc/largeDoc document generators. So, 2 test cases in total. In other words, I don't think we want to test empty documents since they aren't "realistic"

mtandrei (changes requested):
Thanks for picking this up Rui! I have two comments and this should be good to go.

testcases/simple_query.js
lriuui0x0: Thanks I have updated the code.

testcases/simple_query.js
```diff
@@ -24,6 +24,51 @@ if (typeof(tests) !== "object") {
         op: {op: "find", query: {}}
     });
 
+    /**
```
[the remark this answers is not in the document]
lriuui0x0: Thanks I have changed this.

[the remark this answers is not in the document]
mtandrei: Thanks; looks good!

testcases/simple_query.js
```diff
@@ -24,6 +24,32 @@ if (typeof(tests) !== "object") {
...
+     * Setup: Create a large collection of 10k large documents.
+     *
+     * Test: Empty query that returns all documents.
+     */
+    addQueryTestCase({
+        name: "Large10k",
+        tags: ["regression"],
+        nDocs: 10000,
```
mtandrei, line 35: [optional] consider generating these two in a loop like so: 

https://github.com/mongodb/mongo-perf/blob/97e50f99585e0c7bd80c065d9ec9a97dc2579f50/testcases/simple_sort_queries.js#L230-L233

mtandrei (approved):
LGTM Rui with an optional suggestion; thanks!

testcases/simple_query.js
lriuui0x0: Thanks for the tip, done!

lriuui0x0:
evergreen merge

PR #123, opened 2023-02-16
PERF-3797 CWI.insert: add core and regression tags

aligusnet:


alyacb (approved):
LGTM- just one question about the test filters in evergreen.yml. Is `filter2` an "AND" or an "OR"? I'm assuming an OR, because if its an AND, it would only run the tests marked as both core & regression.

RuoxinXu (approved):
LGTM

aligusnet:
Thanks you for the review, Alya. Inside one filter patameter it is "OR", and "AND" between filter parameters.
So let's say we have filter1 = "compound-wildcard-insert" and filter2: "core regression"
it effectively will be `(tags.contains("compound-wildcard-insert")) && (tags.constains("core") || tags.constains("regression"))`

PR #138, opened 2023-09-11
SERVER-80273 Implement tie breaking performance tests

aligusnet:
The test results in op/sec. with the tie breaking knob enabled and disabled. 

```
TieBreaking.Longest Index Prefix
8       513.8479528896214
Regular.Longest Index Prefix
8       44.694536921251505
TieBreaking.Equality
8       506.8080551614268
Regular.Equality
8       161.7573477034724
TieBreaking.Shortest Index
8       510.61094659644266
Regular.Shortest Index
8       429.83690121885473
TieBreaking.Shortest Index With Comparisons
8       241.43640952463656
Regular.Shortest Index With Comparisons
8       237.93900544287968
TieBreaking.Not Broken Tie
8       520.1677700619116
Regular.Not Broken Tie
8       501.3517774484123
TieBreaking.Multi Interval Index Bounds
8       1418.8523158938228
Regular.Multi Interval Index Bounds
8       589.0653669061361
TieBreaking.Non-Blocking Sort
8       1399.3737582655758
Regular.Non-Blocking Sort
8       605.9069944737929
TieBreaking.Blocking Sort
8       1293.7228272758164
Regular.Blocking Sort
8       563.8432802989446
TieBreaking.Multi IndexScans
8       1269.5654745325644
Regular.Multi IndexScans
8       530.3681668955684
TieBreaking.No Tie
8       469.9333380519399
Regular.No Tie
8       487.1500466012959
TieBreaking.Docs Examined
8       2842.6274584899634
Regular.Docs Examined
8       2997.623491292798
```

### DIFF
--- a/testcases/tie_breaking.js
+++ b/testcases/tie_breaking.js
@@ -17,7 +17,7 @@ const docsExaminedDocs = [];
 for (let i = 0; i < 200; ++i) {
     docsExaminedDocs.push({i: i, a: "Jerry", b: "not mouse", c: "Tom", d: "degu"});
 }
-for (let i = 0; i < 1100; i++) {
+for (let i = 0; i < 1100; ++i) {
     docsExaminedDocs.push({i: i, a: "Jerry", b: "mouse", c: "Tom", d: "degu"});
 }
 
--- a/testcases/tie_breaking.js
+++ b/testcases/tie_breaking.js
@@ -1,3 +1,9 @@
+/**
+ * Additional tie breaking heuristics perfomance tests:
+ * - index prefix heuristic
+ * - docsExamined heuristic
+ */
+
 if (typeof (tests) != "object") {
     tests = [];
 }
@@ -7,21 +13,23 @@ if (typeof (tests) != "object") {
 
 // Generate documents for testing the index prefix heuristics.
 const indexPrefixDocs = [];
-for (let i = 0; i < 1000; ++i) {
+for (let i = 0; i < 100000; ++i) {
     indexPrefixDocs.push({a: 1, b: "hello", c: i * 12, d: 111 * i - 100, h: i});
     indexPrefixDocs.push({a: i + 1000, b: `hello%{i}`, c: i * 77, d: -i, h: i});
 }
 
 // Generate documents for testing the number of documents examined heuristics.
 const docsExaminedDocs = [];
+// Adding more than 101 documents to make sure we don't hit EOF.
 for (let i = 0; i < 200; ++i) {
     docsExaminedDocs.push({i: i, a: "Jerry", b: "not mouse", c: "Tom", d: "degu"});
 }
+// Some additional payload data.
 for (let i = 0; i < 1100; ++i) {
     docsExaminedDocs.push({i: i, a: "Jerry", b: "mouse", c: "Tom", d: "degu"});
 }
 
-const baseCases = [
+const perfCases = [
     {
         name: "Longest Index Prefix",
         indexes: [{a: 1}, {b: 1, a: 1}],
@@ -91,19 +99,6 @@ const baseCases = [
         docs: docsExaminedDocs,
     },
 ];
-
-const perfCases = [];
-for (let baseCase of baseCases) {
-    // Create a performance test case with the tie breaking heuristics enabled.
-    const perfCaseTB = Object.assign({enableTieBreaking: true}, baseCase);
-    perfCaseTB.name = `TieBreaking.${perfCaseTB.name}`;
-    perfCases.push(perfCaseTB);
-
-    // Create a performance test case with the tie breaking heuristics disabled.
-    const perfCaseNTB = Object.assign({enableTieBreaking: false}, baseCase);
-    perfCaseNTB.name = `Regular.${perfCaseNTB.name}`;
-    perfCases.push(perfCaseNTB);
-}
 
 // Returns setup function for the given perfomance test case.
 function getSetupFunction(perfCase) {
--- a/testcases/tie_breaking.js
+++ b/testcases/tie_breaking.js
@@ -94,11 +94,14 @@ const baseCases = [
 
 const perfCases = [];
 for (let baseCase of baseCases) {
+    // Create a performance test case with the tie breaking heuristics enabled.
     const perfCaseTB = Object.assign({enableTieBreaking: true}, baseCase);
-    perfCaseTB.name = `TB.${perfCaseTB.name}`;
+    perfCaseTB.name = `TieBreaking.${perfCaseTB.name}`;
     perfCases.push(perfCaseTB);
+
+    // Create a performance test case with the tie breaking heuristics disabled.
     const perfCaseNTB = Object.assign({enableTieBreaking: false}, baseCase);
-    perfCaseNTB.name = `NTB.${perfCaseNTB.name}`;
+    perfCaseNTB.name = `Regular.${perfCaseNTB.name}`;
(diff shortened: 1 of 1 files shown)


testcases/tie_breaking.js
```diff
@@ -0,0 +1,135 @@
...
+            setParameter: 1,
+            internalQueryPlanTieBreakingWithIndexHeuristics: perfCase.enableTieBreaking
+        }));
+        assert.commandWorked(collection.createIndexes(perfCase.indexes));
+        assert.commandWorked(collection.insertMany(perfCase.docs));
+    }
+}
+
+// Create test cases for the benchmark.
+for (let perfCase of perfCases) {
```
alyacb, line 122: nit: const

testcases/tie_breaking.js
```diff
@@ -0,0 +1,135 @@
...
+    {
+        name: "Docs Examined",
+        indexes: [{i: 1, a: 1, c: 1}, {i: 1, a: 1, b: 1}],
+        query: {a: "Jerry", b: /not mouse/, c: /Tom/, d: /degu/, i: {$gt: 10}},
+        docs: docsExaminedDocs,
+    },
+];
+
+const perfCases = [];
+for (let baseCase of baseCases) {
```
alyacb, line 96: nit: const

testcases/tie_breaking.js
```diff
@@ -0,0 +1,135 @@
+if (typeof (tests) != "object") {
+    tests = [];
+}
+
+(function() {
+'use strict';
+
+// Generate documents for testing the index prefix heuristics.
+const indexPrefixDocs = [];
+for (let i = 0; i < 1000; ++i) {
```
alyacb, line 10: I would expect the choice of plan to make a bigger difference for a larger collection- furthermore, since the tests have filters with some selectivity in them, are we sure we're not hitting an EOF plan in any of the cases?

Can we increase this constant to at least 100k?

alyacb (changes requested):
Thanks Alex! I had a look and left a few comments.

I have one additional request: could you do a local run of your workload with/without the query knob enabled, and post the respective throughputs in this PR for posterity? Once this workload is merged, we will only detect regressions/improvements if the code subsequently changes, but we won't have any numbers anywhere that definitively show the tie-breaker was a good idea.

testcases/tie_breaking.js
antkorsh: +1

testcases/tie_breaking.js
```diff
@@ -0,0 +1,135 @@
+if (typeof (tests) != "object") {
```
antkorsh, line 1: Don't we describe the purpose of a perf test in a same way as we do with js tests? I think it's worthwhile to do.

testcases/tie_breaking.js
```diff
@@ -0,0 +1,135 @@
...
+    indexPrefixDocs.push({a: 1, b: "hello", c: i * 12, d: 111 * i - 100, h: i});
+    indexPrefixDocs.push({a: i + 1000, b: `hello%{i}`, c: i * 77, d: -i, h: i});
+}
+
+// Generate documents for testing the number of documents examined heuristics.
+const docsExaminedDocs = [];
+for (let i = 0; i < 200; ++i) {
+    docsExaminedDocs.push({i: i, a: "Jerry", b: "not mouse", c: "Tom", d: "degu"});
+}
+for (let i = 0; i < 1100; ++i) {
```
antkorsh, line 20: Different limits on the number of inserted docs deserves an explanation, IMO.

antkorsh (approved):

testcases/tie_breaking.js
aligusnet: Done

testcases/tie_breaking.js
aligusnet: Makes sense, increased to 100000.

testcases/tie_breaking.js
aligusnet: Added comments.

testcases/tie_breaking.js
aligusnet: We discussed earlier on a previous PR: `const` does not work here, most likely because legacy `mongo` shell with old JS interpreter is used.

testcases/tie_breaking.js
```diff
@@ -0,0 +1,135 @@
...
+    const perfCaseNTB = Object.assign({enableTieBreaking: false}, baseCase);
+    perfCaseNTB.name = `Regular.${perfCaseNTB.name}`;
+    perfCases.push(perfCaseNTB);
+}
+
+// Returns setup function for the given perfomance test case.
+function getSetupFunction(perfCase) {
+    return function(collection) {
+        collection.drop();
+        assert.commandWorked(db.adminCommand({
```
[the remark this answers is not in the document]
aligusnet: Sure, removed that code.

testcases/tie_breaking.js
aligusnet: `const` does not work here as explained above.

aligusnet (commented):
Thank you for your review, Alya and Anton. I have addressed your comments and added the test run results with and without the tie breaking heuristics flag enabled.

testcases/tie_breaking.js
alyacb: Oops! Forgot. Resolving.

testcases/tie_breaking.js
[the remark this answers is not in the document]
alyacb: Did you push the change? It looks like we're still testing two cases, when we should just be testing whatever is enabled in master.

alyacb (approved):
LGTM, mod last comment!

alyacb:
Actually, last last request: can you update the numbers in the PR with the larger document size? The numbers look quite similar in most cases. I'm surprised that the plans would result in such similar throughputs.

testcases/tie_breaking.js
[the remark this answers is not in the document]
aligusnet: Sorry, forgot to remove this particular line, only removed setting `enableTieBreaking` parameter. Fixed now

aligusnet:
@alyacb Do you mean that not all the tests demonstrated performance boost with the heuristics? It really depends not on the document size, but on how the data distributed. I can imagine, that in some cases, e.g. for the query `{a: 1, b: 1, c: 1}` and the indexes `{a: 1, b: 1}` and `{c: 1}` we would prefer `{a: 1, b: 1}`, but `{c: 1}` might be more selective, and instead of the performance improvement we get performance degradation. There are similar issues with CE, sometimes CE can give misleading results and poorer performance.

alyacb:
@aligusnet I'm surprised none of the tests above show any significant difference (>10%) in throughput, except for this one, where tie breaking is doing better, as expected:
```
TieBreaking.Equality
8       53.38707500884039
Regular.Equality
8       21.28536077650503
```

I want to be sure that:
1. we're actually tie breaking/not picking the first EOF plan in this performance test
2. that we design this test to specifically exercise tie breaking in the beneficial case, so that we can detect regressions in this case in future
3.  we have throughput numbers that support this optimization being a positive one, though I agree with your point that it will be worse in some cases, like any heuristic

aligusnet:
@alyacb What I wanted to demonstrate here is that the heuristics did not make the plans worse in the cases when they shouldn't (in other words not too expensive to apply) and that sometimes they can give performance boost.

If you believe that only cases with performance boost count I can remove others.

PR #142, opened 2023-10-19
PERF-4087 Added appropriate dataset sizes for lookup with disabled cache

Fefer-Ivan:
**Jira Ticket:** PERF-4087

**Whats Changed:**
Added Lookup.UncorrelatedJoin.NoCache.LargeCollection and Lookup.UncorrelatedPrefixJoin.NoCache.LargeCollection perf tests. 

**Patch testing results:**
https://spruce.mongodb.com/version/6530faf7d1fe079737233a97/tasks

### DIFF
--- a/testcases/pipelines.js
+++ b/testcases/pipelines.js
@@ -959,7 +959,9 @@ function basicUncorrelatedPipelineLookupPopulator(isView, disableCache, nDocs) {
         return {_id: val};
     }
 
-    const nDocs = nDocs ? nDocs : 50;
+    if (nDocs === undefined) {
+        nDocs = 50;
+    }
     if (disableCache === undefined) {
         disableCache = false;
     }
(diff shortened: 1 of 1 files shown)


mattBoros (approved):
LGTM, just curious if you think 1000 is high enough or we should increase even more.

testcases/pipelines.js
```diff
@@ -1178,53 +1180,14 @@ generateTestCase({
             }
         }
     ],
-    tags: ["lookup", ">=3.5"]
-});
-
+    false /*allowLargeDataset*/
```
Fefer-Ivan, line 1183: So for full uncorrelated join we can't do large dataset because it means that the whole collection > 100 MB which means that we can't put it into the whole document.

When only prefix is uncorrelated, we can do it, because resulting document will be < 16 MB (I have adjusted the size for it).

The result is that yes, NoCache.LargeData and .LargeData are equal

mattBoros: Thanks for looking into this! SGTM

mattBoros (approved):

PR #146, opened 2024-01-04
SERVER-84399 Extend existing projection tests to run with cardinalities

RuoxinXu:
Thanks for submitting a PR to the Mongo-Perf repo. Please include the following fields (if relevant) prior to submitting your PR.

**Jira Ticket:** SERVER-84399

**Whats Changed:**
Extend existing projection tests to run with cardinalities

**Patch testing results:**
System failure patch (hit 6hr timed-out because the 1M cardinality is too big for COLLSCAN plan.): 
https://spruce.mongodb.com/version/659556f6306615d965b84f58/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

### DIFF
--- a/testcases/simple_query.js
+++ b/testcases/simple_query.js
@@ -361,11 +361,15 @@ if (typeof(tests) !== "object") {
      *
      * Test: Query for all documents (empty query) and return the three integer fields.
      */
-    for (const numOfDocs of [10, 1000, 100000]) {
+    for (const testOpts of [{num: 100, name: ""}, {num: 10000, name: "Medium"}, {num: 100000, name: "Large"}]) {
+        let tags = ["regression"];
+        if (testOpts.num >= 100000) {
+            tags.push("query_large_dataset");
+        }
         addQueryTestCase({
-            name: "FindProjectionThreeFields_Cardinality" + numOfDocs,
-            tags: ["regression"],
-            nDocs: numOfDocs,
+            name: "FindProjectionThreeFields" + testOpts.name,
+            tags: tags,
+            nDocs: testOpts.num,
             docs: function(i) {
                 return {x: i, y: i, z: i};
             },
@@ -391,15 +395,19 @@ if (typeof(tests) !== "object") {
 
     /**
      * Utility to add a pair of inclusion/exclusion test cases. These test cases are run against
-     * collections with different cardinalities of 100, 10k, and 1M.
+     * collections with different cardinalities of 100, 10k, and 100K.
      */
     const addInclusionExclusionTestCase = function(name, docGenerator, inclusionSpec, exclusionSpec) {
-        for (const numOfDocs of [10, 1000, 100000]) {
+        for (const testOpts of [{num: 100, name: ""}, {num: 10000, name: "Medium"}, {num: 100000, name: "Large"}]) {
+            let tags = ["regression", "projection", ">=4.4.0"];
+            if (testOpts.num >= 100000) {
+                tags.push("query_large_dataset");
+            }
             for (const [prefix, testCase] of Object.entries({"FindInclusion.": inclusionSpec, "FindExclusion.": exclusionSpec})) {
                 addQueryTestCase({
-                    name: prefix + name + "_Cardinality" + numOfDocs,
-                    tags: ["regression", "projection", ">=4.4.0"],
-                    nDocs: numOfDocs,
+                    name: prefix + name + testOpts.name,
+                    tags: tags,
+                    nDocs: testOpts.num,
                     // Adding a views passthrough and an aggregation test would be redundant.
                     createViewsPassthrough: false,
                     createAggregationTest: false,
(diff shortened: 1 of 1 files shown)


testcases/simple_query.js
```diff
@@ -361,15 +361,17 @@ if (typeof(tests) !== "object") {
...
-        tags: ["regression"],
-        nDocs: 100,
-        docs: function(i) {
-            return {x: i, y: i, z: i};
-        },
-        op: {op: "find", query: {}, filter: {x: 1, y: 1, z: 1, _id: 0}}
-    });
+    for (const numOfDocs of [10, 1000, 100000]) {
+        addQueryTestCase({
+            name: "FindProjectionThreeFields_Cardinality" + numOfDocs,
```
alyacb, line 366: I'd prefer it if we didn't rename the smallest cardinality- if we do, we will lose history for this metric. I'd also prefer if we stuck to the Small/Medium/Large naming convention or similar, rather than including CardinalityX in the name. This is to help make the dashboard more consistent.

I think here we can keep the lowest cardinality as 100 instead of adding 10.

testcases/simple_query.js
```diff
@@ -388,20 +390,23 @@ if (typeof(tests) !== "object") {
...
-                createAggregationTest: false,
-                docs: docGenerator,
-                op: {op: "find", query: {}, filter: testCase}
-            });
+        for (const numOfDocs of [10, 1000, 100000]) {
+            for (const [prefix, testCase] of Object.entries({"FindInclusion.": inclusionSpec, "FindExclusion.": exclusionSpec})) {
+                addQueryTestCase({
+                    name: prefix + name + "_Cardinality" + numOfDocs,
+                    tags: ["regression", "projection", ">=4.4.0"],
+                    nDocs: numOfDocs,
```
alyacb, line 402: Again, here I'd like us to keep the 10k case- could replace the 1k case.

testcases/simple_query.js
```diff
@@ -574,6 +579,29 @@ if (typeof(tests) !== "object") {
         });
     }
 
+    function addTestCaseWithMultipleDatasets(options, cardinalities) {
+        for (const numOfDocs of cardinalities) {
+            let tags = options.tags || [];
+            tests.push({
+                tags: ["regression", "query_large_dataset"].concat(tags),
```
alyacb, line 586: We really should only tag things as "query_large_dataset" if they actually have a large cardinality.

testcases/simple_query.js
```diff
@@ -574,6 +579,29 @@ if (typeof(tests) !== "object") {
         });
     }
 
+    function addTestCaseWithMultipleDatasets(options, cardinalities) {
```
alyacb, line 582: Why not just modify the existing function? As long as we don't delete the old metrics, even if we extend some test we don't care about for M2, we will be saving ourselves work in M4+.

alyacb (changes requested):
Thanks Roy! I left a few comments. Mostly I want to make sure we're not losing any test coverage.

testcases/simple_query.js
```diff
@@ -551,27 +568,37 @@ if (typeof(tests) !== "object") {
...
-                    for (var i = 0; i < nDocs; i++) {
-                        bulkop.insert(options.docGenerator(i));
-                    }
-                    bulkop.execute();
-                },
-            pre: options.pre || function (collection) {},
-            post: options.post || function(collection) {},
-            ops: ("op" in options) ? [options.op] : [{op: "find", query: options.query}],
-        });
+        const nDocs = options.nDocs || largeCollectionSize;
```
alyacb, line 573: I actually might prefer just forcing all tests to use cardinalities. Could we remove the nDocs option, default to an array of [largeCollectionSize] elements for cardinalities if its not specified? It seems like we're hard-coding this value for the tests below anyway.

alyacb (changes requested):
Thanks Roy! I have a couple more comments- thanks for bearing with me 😅

testcases/simple_query.js
```diff
@@ -751,68 +785,76 @@ if (typeof(tests) !== "object") {
...
             createIndexes(collection, options.indexes);
         };
         options.post = dropIndexesAndCaches;
-        addTestCaseWithLargeDataset(options);
+        addTestCaseWithMultipleDatasets(options);
     }
 
-    addTestCaseWithLargeDatasetAndIndexes({
-        name: "PointQuery_SingleIndex_LL",
+    addTestCaseWithMultipleDatasetsAndIndexes({
```
alyacb, line 797: nit: couldn't we create an addTestCaseWithLargeDatasetAndIndexes function which calls addTestCaseWithMultipleDatasets and sets cardinalities? I think this would minimize churn in this PR, but maybe being explicit is clearer, so I'll leave this up to you.

alyacb (approved):
LGTM, thanks Roy!

PR #154, opened 2024-04-04
SERVER-88984: Fix high_value tag for UnindexedVeryLargeInUnsortedMatchingBigCollection

nzolnierzmdb:
Fixes a typo, note the suffix "BigCollection" vs "VeryBigCollection" 😞


HanaPearlman (approved):
So we're just swapping which one is marked as high-value? Makes sense!

PR #158, opened 2024-05-15
SERVER-90613 Zixuan Skunkworks 2024: Improve microbenchmark benchrun.py

leozzx:
1. Includes median and stdev for each test result
2. Allows users to provide a variant name and list of variant values and comparing the perf with different variants.
3. Allow users to use regex for filtering test names
4. Prints a summary at the end of test.

https://spruce.mongodb.com/version/664bdf7ab3280b0007d8c421/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

New `--tsvSummary` option output looks like
```
name                     variant    thread_count    ops_per_sec(mean)    ops_per_sec(median)    count    stdev
---------------------  ---------  --------------  -------------------  ---------------------  -------  -------
Aggregation.Group.All          0               1            1628.9812              1627.5134        3   3.7440
Aggregation.Group.All          1               1            1576.7357              1571.2285        3  17.9333
Aggregation.Group.All          2               1            1620.0208              1606.8993        3  20.7279
Aggregation.Group.All          4               1            1592.5159              1593.8476        3   5.9513
Aggregation.Group.All          8               1            1603.3958              1583.7102        3  29.6256

### DIFF
--- a/util/utils.js
+++ b/util/utils.js
@@ -128,7 +128,7 @@ function CommandTracer(testName, options) {
         runningPre: "running pre() function",
         runningOps: "running benchRun ops",
         runningPost: "running post() function",
-        done: "done",
+        done: "done"
     };
 
     TestData = TestData || {};
@@ -471,6 +471,30 @@ function getMean(values) {
     return sum / values.length;
 }
 
+function getMedian(values) {
+    var sorted = Array.from(values).sort((a, b) => a - b);
+    var middle = Math.floor(sorted.length / 2);
+
+    if (sorted.length % 2 === 0) {
+        return (sorted[middle - 1] + sorted[middle]) / 2;
+    }
+
+    return sorted[middle];
+}
+
+function getStdev(values, mean) {
+    // Assigning (value - mean) ^ 2 to every array item
+    var arr = values.map((k) => {
+        return (k - mean) ** 2
+    });
+
+    // Calculating the sum of updated array
+    let sum = arr.reduce((acc, curr) => acc + curr, 0);
+
+    // Returning the standard deviation
+    return Math.sqrt(sum / arr.length)
+}
+
 function getNFieldNames(numFields) {
     var fieldNames = [];
     for (var i = 0; i < numFields; i++) {
@@ -518,14 +542,16 @@ function doCompare(test, compareTo) {
 
     if ( Array.isArray(compareTo) ) {
         for (var i = 0;i < compareTo.length; i++) {
-            if ( tags.indexOf(compareTo[i]) > -1 || test.name == compareTo[i]) {
+            const regx = new RegExp(compareTo[i] + "$");
+            if ( tags.indexOf(compareTo[i]) > -1 || test.name.match(regx)) {
                 return true;
             }
         }
         return false;
     }
     else {
-        if ( tags.indexOf(compareTo) > -1 || test.name == compareTo) {
+        const regx = new RegExp(compareTo + "$");
+        if ( tags.indexOf(compareTo) > -1 || test.name.match(regx)) {
             return true;
         }
     }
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: util/utils.js [hunk 1]
```diff
@@ -696,59 +825,50 @@ function runTests(
...
+                                                   shareDataset,
+                                                   "",
+                                                   mongoeBenchOptions,
+                                                   username,
+                                                   password,
+                                                   testResults.errors);
+                testResults['results'].push({name: test.name, results: threadResults});
+            } else {
+                for (var variant of variants) {
+                    db.adminCommand({setParameter: 1, [variantName]: NumberLong(variant)});
```
anna-wawrzyniak, line 849: there doesn't seem to be guaranteed cleanup. Ideally we should unset this value under the test, and make sure that it happens even if test fails, or ctrl-c is pressed.

util/utils.js
```diff
@@ -518,14 +539,16 @@ function doCompare(test, compareTo) {
...
+            const regx = new RegExp(compareTo[i] + "$");
+            if ( tags.indexOf(compareTo[i]) > -1 || test.name.match(regx)) {
                 return true;
             }
         }
         return false;
     }
     else {
-        if ( tags.indexOf(compareTo) > -1 || test.name == compareTo) {
+        const regx = new RegExp(compareTo + "$");
```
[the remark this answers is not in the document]
leozzx: This is because if the input is full name of a test, the regex will treat it as prefix without '$' and all the tests starting the name will be ran. Adding '$' will prevent this and if developer want to match a prefix, we can add '.*' to the end of test name.

util/utils.js
leozzx: Suppose this is a temporary variant developer add for perf testing, it is ok not to cleanup, but I can restore it after the test.

util/utils.js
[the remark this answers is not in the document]
anna-wawrzyniak: is seems rather arbitrary that we will explicitly not match prefixes, but will match suffixes. It seems like it should be either no modification, or doing: "^" + compareTo + "$" and require that partial matches should use .* on the both sides.

util/utils.js
anna-wawrzyniak: even for temporary variant. If I run:
```benchrun.py --variantName="foo" --variants="0 1 2" stuff```
and then run
```benchrun.py stuff```

I'd expect that the 1st run will not leave side effects behind, and that my second benchrun.py will run with regular settings (no variant knobs set).

File: util/utils.js [hunk 3]
```diff
@@ -768,14 +892,50 @@ function runTests(
...
+                           excludeTestbed,
+                           printArgs,
+                           shareDataset,
+                           variantName,
+                           variants,
+                           mongoeBenchOptions,
+                           username,
+                           password) {
+    var testResults = "";
+    var oldValue = db.adminCommand({getParameter: 1, [variantName]: 1});
```
anna-wawrzyniak, line 917: does this work if variantName is not set?

leozzx: Thanks for catching it. It fails silently, but good to prevent it.

util/utils.js
```diff
@@ -768,14 +892,52 @@ function runTests(
...
+                           shareDataset,
+                           variantName,
+                           variants,
+                           mongoeBenchOptions,
+                           username,
+                           password) {
+    var testResults = "";
+    var oldValue = variantName === null ? null : db.adminCommand({getParameter: 1, [variantName]: 1});
+    try {
+        testResults = runTests(threadCounts,
```
anna-wawrzyniak, line 919: nit: is there any way this argument list can be packages into object/struct so that we don't have to carry lists of 20 args  around?

anna-wawrzyniak (approved):

anna-wawrzyniak:
The new json output is backwards compatible with the previous one, right?

